PR #1, opened 2021-02-10
Added shape_active condition

Arrinao:
Added shape_active condition but the program keeps returning an error.


Akuli:
I talked with Arrinao and we decided to close this

Issue #3, opened 2021-02-11
clean up new_block() function


Akuli:
- [x] last 3 lines of each `if` and `elif` are the same, get rid of copy/pasta by putting that once after the last `elif`
- [x] rename global `x` and `y` variables to e.g. `spawn_x` and `spawn_y` to avoid mixing up with local `x` and `y` variables

PR #4, opened 2021-02-11
fix copy/pasta

Akuli:
checks a checkbox in #3


Akuli:
someone else fixed this

PR #15, opened 2021-02-17
Added full_line_clear function

Arrinao:
After I finally figured out how collections.Counter works, I decided to revisit my original idea of the function today. The result seems to be working fine now. I spent good 1,5 hours on the last line before going back to your code and noticing that you actually also added a condition for (x, y) BELOW the y_line. Only after adding this condition (and creating that nasty 100 mile long line) I made the whole thing to work. I still don't quite understand the mechanism behind it as I still don't see why would the code for (x, y) above the y_line not be enough. See you this evening.

### DIFF
--- a/tetris.py
+++ b/tetris.py
@@ -202,10 +202,22 @@ class TetrisGame:
         print(coordinates_counter)
         for y_line, count in coordinates_counter.items():
             if count == game_width:
-                self.landed_blocks = [(a, b) for (a, b) in self.landed_blocks if b != y_line]
-                time.sleep(0.5)
+                #root.after() here
                 self.landed_blocks = [(a, b+1) for (a, b) in self.landed_blocks if b < y_line] + [(a, b) for (a, b) in self.landed_blocks if b > y_line]
             
 
+            
+    #def full_line_clear(self):
+    #    counter = collections.Counter([y for x, y in self.landed_blocks])
+    #    print(counter)
+    #    for line_y in range(game_height):
+    #        if counter[line_y] == game_width:
+    #            new_landed_blocks = []
+    #            for x, y in self.landed_blocks:
+    #                if (x, y)[1] < line_y:
+    #                    new_landed_blocks.append((x, y+1))   # move down
+    #                elif (x, y)[1] > line_y:
+    #                    new_landed_blocks.append((x, y))
+    #            self.landed_blocks = new_landed_blocks
 run_gui()
 
--- a/tetris.py
+++ b/tetris.py
@@ -207,17 +207,4 @@ class TetrisGame:
                 self.landed_blocks = [(a, b+1) for (a, b) in self.landed_blocks if b < y_line] + [(a, b) for (a, b) in self.landed_blocks if b > y_line]
             
 
-            
-    #def full_line_clear(self):
-    #    counter = collections.Counter([y for x, y in self.landed_blocks])
-    #    print(counter)
-    #    for line_y in range(game_height):
-    #        if counter[line_y] == game_width:
-    #            new_landed_blocks = []
-    #            for x, y in self.landed_blocks:
-    #                if (x, y)[1] < line_y:
-    #                    new_landed_blocks.append((x, y+1))   # move down
-    #                elif (x, y)[1] > line_y:
-    #                    new_landed_blocks.append((x, y))
-    #            self.landed_blocks = new_landed_blocks
 run_gui()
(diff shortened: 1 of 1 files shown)


tetris.py
```diff
@@ -192,5 +196,16 @@ def block_rotator(self, event):
             rotate.append((y, x))
         self.current_block = rotate
 
+    def full_line_clear(self):
+        y_coordinates = [y for (x, y) in self.landed_blocks]
+        coordinates_counter = collections.Counter(y_coordinates)
+        print(coordinates_counter)
+        for y_line, count in coordinates_counter.items():
```
Akuli, line 203: The counter is not necessarily in the correct order:

```
>>> collections.Counter([(0, 0), (0, 1)])
Counter({(0, 0): 1, (0, 1): 1})
>>> collections.Counter([(0, 1), (0, 0)])
Counter({(0, 1): 1, (0, 0): 1})
```

Notice how in these examples, the coordinate with bigger `y` can be first or second in the counter, depending on the order of the list passed in. In this case, the order of the list depends on how the blocks happen to fall down, and can't be controlled.

To fix this, you can loop over `range(height)` which is always in increasing order, or `reversed(range(height))` which is always in decreasing order:

```suggestion
        for y_line in range(game_height):   # TODO: should this be reversed?
            count = coordinates_counter[y_line]
```

tetris.py
```diff
@@ -192,5 +196,16 @@ def block_rotator(self, event):
...
+            if count == game_width:
+                self.landed_blocks = [(a, b) for (a, b) in self.landed_blocks if b != y_line]
+                time.sleep(0.5)
```
Akuli, line 206: Please don't use `time.sleep` in tkinter programs. It freezes the gui and does not improve anything. Delete this line.

tetris.py
```diff
@@ -192,5 +196,16 @@ def block_rotator(self, event):
             rotate.append((y, x))
         self.current_block = rotate
 
+    def full_line_clear(self):
+        y_coordinates = [y for (x, y) in self.landed_blocks]
+        coordinates_counter = collections.Counter(y_coordinates)
+        print(coordinates_counter)
+        for y_line, count in coordinates_counter.items():
+            if count == game_width:
+                self.landed_blocks = [(a, b) for (a, b) in self.landed_blocks if b != y_line]
```
Akuli, line 205: This lines deletes the blocks on the `y_line`, but that's not necessary, because the 1km line below deletes them too (it does not delete blocks with `b > y_line` or `b < y_line`, and deletes other blocks, those with `b == y_line`).

Akuli (changes requested):
Nice work! The code looks good overall, even though it still needs a couple small changes.

The 1km long line is fine, because `black` will split it to multiple lines and make it readable again. You need the second part for it, because the first part deletes all blocks below `y_line` when it does `if b < y_line`; the second part adds those back.

tetris.py
Akuli: I tried playing the game with the `time.sleep`. It actually works quite fine. It's not the cleanest solution, but it works.

tetris.py
Akuli: Sometimes the order happens to be correct, and multiple full lines are deleted properly. But sometimes the order is wrong and only one full line is deleted.

tetris.py
```diff
@@ -192,5 +196,28 @@ def block_rotator(self, event):
...
 
-
+    def full_line_clear(self):
+        y_coordinates = [y for (x, y) in self.landed_blocks]
+        coordinates_counter = collections.Counter(y_coordinates)
+        print(coordinates_counter)
+        for y_line in range(game_height):   # TODO: should this be reversed?
+            count = coordinates_counter[y_line]
+            if count == game_width:
+                #root.after() here
```
Akuli, line 206:
```suggestion
                # TODO: root.after() here
```

tetris.py
```diff
@@ -192,5 +196,28 @@ def block_rotator(self, event):
...
+    #    print(counter)
+    #    for line_y in range(game_height):
+    #        if counter[line_y] == game_width:
+    #            new_landed_blocks = []
+    #            for x, y in self.landed_blocks:
+    #                if (x, y)[1] < line_y:
+    #                    new_landed_blocks.append((x, y+1))   # move down
+    #                elif (x, y)[1] > line_y:
+    #                    new_landed_blocks.append((x, y))
+    #            self.landed_blocks = new_landed_blocks
```
Akuli, line 222:
```suggestion
```

Akuli (changes requested):

PR #19, opened 2021-02-18
Square color

Arrinao:


tetris.py
```diff
@@ -172,7 +189,8 @@ def block_mover(self):
     def block_rotator(self, event):
         rotate = []
         for (x, y) in self.current_block:
-            rotate.append((y, x))
+            if self.current_block == 'I':
+                rotate.append((y, x))
```
Akuli, line 193: This deletes all blocks except `I` blocks. Don't try to fix this function in the same pull request.

tetris.py
```diff
@@ -132,6 +144,10 @@ def new_block(self):
             self.current_block = self.upcoming_block
         self.upcoming_block = random.choice(list(blocks.values()))
 
+    def square_color(self):
+        k = {(coord, random.choice(list(color_chart.values()))) for coord in self.landed_blocks}
+        print(k)
```
Akuli, line 149: How is this method supposed to work? It doesn't return anything, and it doesn't change any attribute of `self`.

Also, it makes a **set**, not a dict. Did you know this? Do you know what a set is?
```
{thing1: thing2 for something in something}    # makes a dict
{(thing1, thing2) for something in something}  # makes a set of tuples
```

tetris.py
```diff
@@ -104,6 +106,16 @@ def right_mediator(self, event):
...
+    3: "yellow",
+    4: "gold",
+    5: "orange",
+    6: "dark orange",
+    7: "red2",
+    8: "red4",
+}
 class TetrisGame:
     def __init__(self):
         self.landed_blocks = []
```
Akuli, line 121: Change this to dict with coords as keys and colors as values:

```suggestion
        self.landed_blocks = {}
```

Of course, you need to change all code that uses `landed_blocks`.

tetris.py
```diff
@@ -132,6 +144,10 @@ def new_block(self):
             self.current_block = self.upcoming_block
         self.upcoming_block = random.choice(list(blocks.values()))
```
Akuli, line 145: In `new_block()`, blocks can't be just lists of coordinates anymore. The shape letter, such as `"L"`, has to be stored into a block too. I'm imagining something like this:

```suggestion
        shape_letter = random.choice(list(blocks.keys()))
        self.upcoming_block = {coord: shape_letter for coord in blocks[shape_letter]}
```

For example, if `shape_letter` happens to be `"L"`, this code would set `self.upcoming_blocks` to

```
{(x - 1, y): "L", (x, y): "L", (x + 1, y): "L", (x + 1, y + 1): "L"}
```
instead of just
```
[(x - 1, y), (x, y), (x + 1, y), (x + 1, y + 1)]
```

tetris.py
```diff
@@ -104,6 +106,16 @@ def right_mediator(self, event):
...
 
+color_chart = {
+    1: "turquoise",
+    2: "yellow green",
+    3: "yellow",
+    4: "gold",
+    5: "orange",
+    6: "dark orange",
+    7: "red2",
+    8: "red4",
```
Akuli, line 117: Maybe the keys should be tetris block shape names, e.g. `"L_rev": "red2"` instead of `7: "red2"`

tetris.py
```diff
@@ -29,6 +30,7 @@ def run_gui():
 
     tetris_gui = TetrisGUI(game_speed, tetris_canvas)
     tetris_gui.tetris_game.new_block()
+    tetris_gui.tetris_game.full_line_clear()
```
Akuli, line 33: This is not necessary. Delete this (and to avoid getting things like this into your pull requests, pull on main before making a new branch).

Akuli (changes requested):

tetris.py
Akuli: Actually, this doesn't change what the program does at all: `self.current_block == 'I'` is always false, because `self.current_block` is not a string.

tetris.py
```diff
@@ -245,5 +259,11 @@ def full_line_clear(self):
                     (a, b + 1) for (a, b) in self.landed_blocks if b < y_line
                 ] + [(a, b) for (a, b) in self.landed_blocks if b > y_line]
 
+    def block_color(self):
+        color_list = [RED, BLUE, GREEN, YELLOW, PURPLE, ORANGE, PINK, TEAL]
+        color_chart = {shape: color for shape, color in zip(shape_names, color_list)}
```
Akuli, line 264:
```suggestion
        color_chart = {'I': RED, 'L': BLUE, 'L_rev': GREEN, 'O': YELLOW, 'E': PURPLE, 'Z': ORANGE, 'Z_rev': PINK}
```

tetris.py
```diff
@@ -245,5 +259,11 @@ def full_line_clear(self):
...
+        return color_chart[self.current_block_shape]
```
Akuli, line 265:
```suggestion
        return color_chart[shape]
```

tetris.py
```diff
@@ -245,5 +259,11 @@ def full_line_clear(self):
                     (a, b + 1) for (a, b) in self.landed_blocks if b < y_line
                 ] + [(a, b) for (a, b) in self.landed_blocks if b > y_line]
 
+    def block_color(self):
```
Akuli, line 262: If we make this take the shape to use as an argument, we can use the same method for coloring landed blocks, once their shapes are remembered.

```suggestion
    def block_color(self, shape):
```

tetris.py
```diff
@@ -80,15 +83,26 @@ def draw_block(self):
...
+            self.tetris_game.get_current_block()
         ):
             self.canvas.create_rectangle(
                 x * square_size,
                 y * square_size,
                 x * square_size + square_size,
                 y * square_size + square_size,
                 tags="block",
-                fill=RED,
+                fill=self.tetris_game.block_color(),
```
Akuli, line 94:
```suggestion
                fill=self.tetris_game.block_color(self.tetris_game.current_block_shape),
```

tetris.py
```diff
@@ -80,15 +83,26 @@ def draw_block(self):
...
+        for x, y in (
+            self.tetris_game.landed_blocks
+        ):
+            self.canvas.create_rectangle(
+                x * square_size,
+                y * square_size,
+                x * square_size + square_size,
+                y * square_size + square_size,
+                tags="block",
+                fill=self.tetris_game.block_color(),
```
Akuli, line 105:
```suggestion
                fill=self.tetris_game.block_color("L"),  # TODO: use correct color
```

Akuli (changes requested):

tetris.py
```diff
@@ -180,12 +185,36 @@ def get_current_block(self):
...
+            self.new_block()
+        else:
+            x, y = self.current_block_center
+            self.current_block_center = (x, y + 1)
+
+    def coord_extractor(self):
+        for block in self.landed_blocks.values():
+            for coord in block:
+                self.coords.append(coord)
+        return self.coords
```
Akuli, line 210:
```suggestion
        coords = []
        for block in self.landed_blocks.values():
            for coord in block:
                coords.append(coord)
        return coords
```

tetris.py
```diff
@@ -180,12 +185,36 @@ def get_current_block(self):
...
+
     def user_input_left(self):
         """
         Moves the current block to the left on the canvas
         """
         if any(x == 0 for (x, y) in self.get_current_block()) or any(
-            (x - 1, y) in self.landed_blocks for x, y in self.get_current_block()
+            (x - 1, y) in self.landed_blocks.values() for x, y in self.get_current_block()
```
Akuli, line 217:
```suggestion
            (x - 1, y) in self.coord_extractor() for x, y in self.get_current_block()
```

tetris.py
```diff
@@ -196,27 +225,12 @@ def user_input_right(self):
         Moves the current block to the right on the canvas
         """
         if any(x == game_width - 1 for x, y in self.get_current_block()) or any(
-            (x + 1, y) in self.landed_blocks for x, y in self.get_current_block()
+            (x + 1, y) in self.landed_blocks.values() for x, y in self.get_current_block()
```
Akuli, line 228: Same here, needs `coord_extractor`

tetris.py
```diff
@@ -226,7 +240,7 @@ def block_rotator(self):
         #    (x, y) in self.landed_blocks for x, y in self.get_current_block()
         # ):
         if any(
-            x not in range(game_width) or (x, y) in self.landed_blocks
+            x not in range(game_width) or (x, y) in self.landed_blocks.values()
```
Akuli, line 243: And same here

Akuli (changes requested):

tetris.py
```diff
@@ -131,6 +134,7 @@ def new_block(self):
         self.current_block_center = (int(game_width / 2), -2)
         self.upcoming_block_shape = random.choice(shape_names)
         self.rotate_counter = 0
+        self.current_block_active = True
```
Akuli, line 137: Delete this:

```suggestion
```

tetris.py
```diff
@@ -80,15 +83,15 @@ def draw_block(self):
 
         self.canvas.delete("block")
         for x, y in (
-            self.tetris_game.get_current_block() + self.tetris_game.landed_blocks
+            self.tetris_game.get_current_block() + list(self.tetris_game.coord_extractor())
```
Akuli, line 86:
```suggestion
            self.tetris_game.get_current_block() + self.tetris_game.coord_extractor()
```

tetris.py
```diff
@@ -180,12 +184,37 @@ def get_current_block(self):
...
+            for coord in self.get_current_block():
+#               self.landed_blocks[self.current_block_shape]= coord   ##Doesn't work, only one coord is created :(
+                self.landed_blocks[self.current_block_shape].append(coord)
+            # self.full_line_clear()
+            self.new_block()
+        else:
+            x, y = self.current_block_center
+            self.current_block_center = (x, y + 1)
+
+    def coord_extractor(self):
```
Akuli, line 205: I think this should be renamed. How about `get_landed_coords`?

Akuli (changes requested):

tetris.py
```diff
@@ -235,15 +248,23 @@ def full_line_clear(self):
...
+    def full_line_clear(self):
+        """
+        Clears the line once it's fully populated with blocks
+        """
+        y_coordinates = [y for (x, y) in self.coord_extractor()]
         coordinates_counter = collections.Counter(y_coordinates)
-        for y_line in range(game_height):
-            count = coordinates_counter[y_line]
+        for x_line in range(game_height):
+            count = coordinates_counter[x_line]
```
Akuli, line 264: `x_line` is a terrible variable name, because the value of this variable is an `y` coordinate.

tetris.py
```diff
@@ -235,15 +248,23 @@ def full_line_clear(self):
...
             if count == game_width:
                 # TODO: root.after() here
-                self.landed_blocks = [
-                    (a, b + 1) for (a, b) in self.landed_blocks if b < y_line
-                ] + [(a, b) for (a, b) in self.landed_blocks if b > y_line]
-
-
+                self.coords = [
+                    (a, b + 1) for (a, b) in self.coord_extractor() if b < x_line
+                ] + [(a, b) for (a, b) in self.coord_extractor() if b > x_line]
```
Akuli, line 269: Please do this, and get rid of the cursed `self.coords`:

```python3
for key in self.landed_blocks.keys():
    self.landed_blocks[key] = [(a, b+1) for (a, b) in self.landed_blocks[key] if b < x_line] + [...]
```

Akuli (changes requested):

Akuli:
You could define a global variable like this:

```python
color_dict = {"L": YELLOW, "I": RED, ...}
```

Then draw like this:

```python
for shape_letter, coords in self.tetris_game.landed_blocks.items():
    color = color_dict[shape_letter]
    for x, y in coords:
        draw the square using x, y and color
```

tetris.py
```diff
@@ -80,17 +85,29 @@ def draw_block(self):
...
+            color = color_dict[shape_letter]
+            for x, y in coords:
+                self.canvas.create_rectangle(
+                    x * square_size,
+                    y * square_size,
+                    x * square_size + square_size,
+                    y * square_size + square_size,
+                    tags="block",
+                    fill=color
+                )
```
Akuli, line 109: This is a bit copy/paste. If it bothers you, you can move the long `self.canvas.create_rectangle(...)` into a new function that takes `x`, `y` and `color` as arguments.

tetris.py
```diff
@@ -115,7 +132,7 @@ def rotate_mediator(self, event):
 
 class TetrisGame:
     def __init__(self):
-        self.landed_blocks = []
+        self.landed_blocks = {}
```
Akuli, line 135:
```suggestion
        self.landed_blocks = {}   # e.g. {"L": [(1, 2), (3, 4)]}
```

tetris.py
```diff
@@ -175,12 +192,38 @@ def get_current_block(self):
...
+        """
+        Moves the current block downwards one square on the canvas
+        """
+        if any(
+            (x, y + 1) in self.coord_extractor() for (x, y) in self.get_current_block()
+        ) or any(y + 1 == game_height for (x, y) in self.get_current_block()):
+            if self.current_block_shape not in self.landed_blocks:
+                self.landed_blocks[self.current_block_shape] = []
+            for coord in self.get_current_block():
+#               self.landed_blocks[self.current_block_shape]= coord   ##Doesn't work, only one coord is created :(
```
Akuli, line 205:
```suggestion
```

tetris.py
```diff
@@ -175,12 +192,38 @@ def get_current_block(self):
...
+                self.landed_blocks[self.current_block_shape].append(coord)
+            self.full_line_clear()
+            print(self.landed_blocks)
```
Akuli, line 208:
```suggestion
```

tetris.py
```diff
@@ -223,24 +251,23 @@ def block_rotator(self):
...
+        for x_line in range(game_height):
+            count = coordinates_counter[x_line]
             if count == game_width:
                 # TODO: root.after() here
-                self.landed_blocks = [
-                    (a, b + 1) for (a, b) in self.landed_blocks if b < y_line
-                ] + [(a, b) for (a, b) in self.landed_blocks if b > y_line]
-
+                for key in self.landed_blocks.keys():
+                    self.landed_blocks[key] = [(a, b+1) for (a, b) in self.landed_blocks[key] if b < x_line] + [(a, b) for (a, b) in self.landed_blocks[key] if b > x_line]
```
Akuli, line 271: Desperately needs blacking

Akuli (changes requested):

Akuli:
Replaced by #32

PR #21, opened 2021-02-20
current_block --> center and shape

Arrinao:


tetris.py
```diff
@@ -127,10 +127,27 @@ def new_block(self):
...
         }
         if self.upcoming_block is None:
-            self.current_block = random.choice(list(blocks.values()))
+            self.current_block_shape = random.choice(list(self.blocks.keys()))
+            self.current_block_center = (x, y)
         else:
             self.current_block = self.upcoming_block
-        self.upcoming_block = random.choice(list(blocks.values()))
+        self.upcoming_block = random.choice(list(self.blocks.keys()))
+        self.current_block = self.get_current_block()
```
Akuli, line 135: Instead of setting the current block to `self.current_block`, replace `.current_block` with `.get_current_block()` everywhere, and fix any problems that occur.

tetris.py
```diff
@@ -127,10 +127,27 @@ def new_block(self):
             "I": [(x - 2, y), (x - 1, y), (x, y), (x + 1, y)],
         }
```
Akuli, line 128: I wish I could do a really long multiline suggestion, but I can't:
- Delete `self.blocks`
- Create a new attribute, `self.shape_names` maybe, with the value `["L", "L_rev", "O", "E", "Z", "Z_rev"]`
- Replace `list(self.blocks.keys()) with `self.shape_names` everywhere

Currently there same lists are copy/pasted in `self.blocks` and `get_current_block`. This allows them to be only in `get_current_block`.

Also, you could use a global variable instead of `self.shape_names`. The list of shape names might be useful in GUI code too.

tetris.py
```diff
@@ -127,10 +127,27 @@ def new_block(self):
...
         if self.upcoming_block is None:
-            self.current_block = random.choice(list(blocks.values()))
+            self.current_block_shape = random.choice(list(self.blocks.keys()))
+            self.current_block_center = (x, y)
         else:
             self.current_block = self.upcoming_block
```
Akuli, line 133:
```suggestion
            self.current_block_shape = self.upcoming_block_shape
            self.current_block_center = (x, y)
```

tetris.py
```diff
@@ -127,10 +127,27 @@ def new_block(self):
             "I": [(x - 2, y), (x - 1, y), (x, y), (x + 1, y)],
         }
         if self.upcoming_block is None:
```
Akuli, line 129: Change `upcoming_block` to `upcoming_block_shape`, because only the shape of the next block is meaningful, not coordinates:

```suggestion
        if self.upcoming_block_shape is None:
```

tetris.py
```diff
@@ -127,10 +127,27 @@ def new_block(self):
...
-            self.current_block = random.choice(list(blocks.values()))
+            self.current_block_shape = random.choice(list(self.blocks.keys()))
+            self.current_block_center = (x, y)
         else:
             self.current_block = self.upcoming_block
-        self.upcoming_block = random.choice(list(blocks.values()))
+        self.upcoming_block = random.choice(list(self.blocks.keys()))
```
Akuli, line 134:
```suggestion
        self.upcoming_block_shape = random.choice(list(self.blocks.keys()))
```

Akuli (changes requested):

tetris.py
Arrinao: The suggestion were part of the new push, I made today, sorry for that

tetris.py
```diff
@@ -117,20 +116,30 @@ def new_block(self):
...
-        }
-        if self.upcoming_block is None:
-            self.current_block = random.choice(list(blocks.values()))
+        if self.upcoming_block_shape is None:
+            self.current_block_shape = random.choice(shapes_name)
+            self.current_block_center = (x, y)
         else:
-            self.current_block = self.upcoming_block
-        self.upcoming_block = random.choice(list(blocks.values()))
+            self.current_block = self.upcoming_block_shape
```
Akuli, line 123:
```suggestion
            self.current_block_shape = self.upcoming_block_shape
            self.current_block_center = (x, y)
```

Akuli (changes requested):

tetris.py
```diff
@@ -12,7 +12,7 @@
 GREEN = "#05ff00"
 GREY = "#666666"
 D_GREY = "#383838"
-
+shapes_name = ['I', 'L', 'L_rev', 'O', 'E', 'Z', 'Z_rev']
```
Akuli, line 15:
```suggestion
shape_names = ['I', 'L', 'L_rev', 'O', 'E', 'Z', 'Z_rev']
```

tetris.py
```diff
@@ -107,8 +107,7 @@ def right_mediator(self, event):
 class TetrisGame:
     def __init__(self):
         self.landed_blocks = []
-        self.current_block = None
-        self.upcoming_block = None
+        self.upcoming_block_shape = None
```
Akuli, line 110: Defining all attributes in `__init__` makes it easy to see what attributes the class has:

```suggestion
        self.upcoming_block_shape = None
        self.upcoming_block_center = None
        self.current_block_center = None
```

tetris.py
```diff
@@ -117,59 +116,66 @@ def new_block(self):
...
-        if self.upcoming_block is None:
-            self.current_block = random.choice(list(blocks.values()))
+        if self.upcoming_block_shape is None:
+            self.current_block_shape = random.choice(shapes_name)
+            self.current_block_center = (x, y)
         else:
-            self.current_block = self.upcoming_block
-        self.upcoming_block = random.choice(list(blocks.values()))
+            self.current_block_shape = self.upcoming_block_shape
+            self.current_block_center = (x, y)
```
Akuli, line 124: The center is set in both cases, so it can be moved outside the `if`,`else`. In other words, do this:

```python3
if blah blah:
    self.current_block_shape = ...
else:
    self.current_block_shape = ...
self.current_block_center = (x, y)
```

Can't do this as a suggestion for whatever reason

tetris.py
```diff
@@ -117,59 +116,66 @@ def new_block(self):
...
+        self.upcoming_block_shape = random.choice(shapes_name)
```
Akuli, line 125:
```suggestion
        self.upcoming_block_shape = random.choice(shape_names)
```

tetris.py
```diff
@@ -117,59 +116,66 @@ def new_block(self):
...
         Moves the current block to the left on the canvas
         """
-        left = []
-        for (x, y) in self.current_block:
-            if x == 0:
-                return
-            left.append((x - 1, y))
-        self.current_block = left
+        x, y = self.current_block_center
+        if x == 0:
```
Akuli, line 149: This doesn't work correctly. Instead of checking whether the center is at the very left edge of the game, it should check whether the x coordinate of any square is at the left edge of the game:

```suggestion
        if any(x == 0 for x, y in self.get_current_block()):
```

tetris.py
```diff
@@ -117,59 +116,66 @@ def new_block(self):
...
         Moves the current block to the right on the canvas
         """
-        right = []
-        for (x, y) in self.current_block:
-            if x == 9:
-                return
-            right.append((x + 1, y))
-        self.current_block = right
+        x, y = self.current_block_center
+        if x == 9:
```
Akuli, line 158:
```suggestion
        if any(x == game_width-1 for x, y in self.get_current_block()):
```

tetris.py
```diff
@@ -117,59 +116,66 @@ def new_block(self):
...
         else:
-            self.current_block = [(x, y + 1) for x, y in self.current_block]
+            x, y = self.current_block_center
+            self.current_block_center = (x, y + 1)
 
-    def block_rotator(self, event):
+    def block_rotator(self):
         rotate = []
         for (x, y) in self.current_block:
             rotate.append((y, x))
```
Akuli, line 181:
```suggestion
    def block_rotator(self, event):
        print("TODO: should rotate")
```

Akuli (changes requested):

tetris.py
```diff
@@ -117,62 +118,67 @@ def new_block(self):
...
-            self.current_block = [(x, y + 1) for x, y in self.current_block]
+            x, y = self.current_block_center
+            self.current_block_center = (x, y + 1)
 
     def block_rotator(self, event):
-        rotate = []
-        for (x, y) in self.current_block:
-            rotate.append((y, x))
+        print("TODO: should rotate")
         self.current_block = rotate
```
Akuli, line 182:
```suggestion
```

tetris.py
```diff
@@ -117,62 +118,67 @@ def new_block(self):
...
         """
-        right = []
-        for (x, y) in self.current_block:
-            if x == 9:
-                return
-            right.append((x + 1, y))
-        self.current_block = right
+        x, y = self.current_block_center
+        if any(x == game_width-1 for x, y in self.get_current_block()):
+            return
```
Akuli, line 161:
```suggestion
        if any(x == game_width-1 for x, y in self.get_current_block()):
            return
        x, y = self.current_block_center
```

tetris.py
```diff
@@ -117,62 +118,67 @@ def new_block(self):
...
-        left = []
-        for (x, y) in self.current_block:
-            if x == 0:
-                return
-            left.append((x - 1, y))
-        self.current_block = left
+        x, y = self.current_block_center
+        if any(x == 0 for x, y in self.get_current_block()):
+            return
+        self.current_block_center = (x-1, y)
```
Akuli, line 153:
```suggestion
        if any(x == 0 for x, y in self.get_current_block()):
            return
        x, y = self.current_block_center
        self.current_block_center = (x-1, y)
```

Akuli (changes requested):

PR #25, opened 2021-02-23
Really new rotate

### LINKED ISSUE #12: Prevent blocks from moving into landed blocks from the side
At the moment you can push blocks into already landed blocks by pressing left or right button

### LINKED ISSUE #13: rotating doesn't work
It does something, but it's not what I would call a rotation

### LINKED ISSUE #20: rotating is laggy
squares aren't redrawn after rotating

Arrinao:
Fixes #13
Fixes #20 
Fixes #12

### DIFF
--- a/tetris.py
+++ b/tetris.py
@@ -120,7 +120,8 @@ class TetrisGame:
             self.current_block_shape = random.choice(shapes_name)
             self.current_block_center = (x, y)
         else:
-            self.current_block = self.upcoming_block_shape
+            self.current_block_shape = self.upcoming_block_shape
+            self.current_block_center = (x, y)
         self.upcoming_block_shape = random.choice(shapes_name)
         self.current_block = self.get_current_block()
 
--- a/tetris.py
+++ b/tetris.py
@@ -12,6 +12,9 @@ RED = "#ff1700"
 GREEN = "#05ff00"
 GREY = "#666666"
 D_GREY = "#383838"
+YELLOW = '#ffd343'
+PURPLE = '7e1e9c'
+
 shape_names = ["I", "L", "L_rev", "O", "E", "Z", "Z_rev"]
 
 
@@ -33,7 +36,7 @@ def run_gui():
 
     root.bind("<Left>", tetris_gui.left_mediator)
     root.bind("<Right>", tetris_gui.right_mediator)
-    root.bind("<Up>", tetris_gui.tetris_game.block_rotator)
+    root.bind("<Up>", tetris_gui.rotate_mediator)
 
     tetris_gui.draw_board()
     tetris_gui.draw_block()
@@ -106,6 +109,10 @@ class TetrisGUI:
         self.tetris_game.user_input_right()
         self.draw_block()
 
+    def rotate_mediator(self, event):
+        self.tetris_game.block_rotator()
+        print('kurva')
+        self.draw_block()
 
 class TetrisGame:
     def __init__(self):
@@ -179,7 +186,7 @@ class TetrisGame:
         """
         Moves the current block to the left on the canvas
         """
-        if any(x == 0 for x, y in self.get_current_block()):
+        if any(x == 0 for (x, y) in self.get_current_block()) or any((x-1, y) in self.landed_blocks for x, y in self.get_current_block()):
             return
         x, y = self.current_block_center
         self.current_block_center = (x - 1, y)
@@ -188,7 +195,7 @@ class TetrisGame:
         """
         Moves the current block to the right on the canvas
         """
-        if any(x == game_width - 1 for x, y in self.get_current_block()):
+        if any(x == game_width - 1 for x, y in self.get_current_block()) or any((x+1, y) in self.landed_blocks for x, y in self.get_current_block()):
             return
         x, y = self.current_block_center
         self.current_block_center = (x + 1, y)
@@ -209,8 +216,13 @@ class TetrisGame:
             x, y = self.current_block_center
             self.current_block_center = (x, y + 1)
 
-    def block_rotator(self, event):
+    def block_rotator(self):
+        '''
+        Rotates the current block
+        '''
         self.index += 1
+        return self.get_current_block()
+        
 
     def full_line_clear(self):
         y_coordinates = [y for (x, y) in self.landed_blocks]
@@ -223,6 +235,7 @@ class TetrisGame:
                 self.landed_blocks = [
                     (a, b + 1) for (a, b) in self.landed_blocks if b < y_line
                 ] + [(a, b) for (a, b) in self.landed_blocks if b > y_line]
+
 
 
 run_gui()
--- a/tetris.py
+++ b/tetris.py
@@ -187,21 +187,6 @@ class TetrisGame:
 
     def block_rotator(self, event):
         self.index += 1
-        self.index %= 4
-        if self.current_block_shape == "I":
-            return self.get_current_block[self.index]
-        if self.current_block_shape == "L":
-            return self.get_current_block[self.index]
-        if self.current_block_shape == "L_rev":
-            return self.get_current_block[self.index]
-        if self.current_block_shape == "O":
-            return
-        if self.current_block_shape == "E":
-            return self.get_current_block[self.index]
-        if self.current_block_shape == "Z":
-            return self.get_current_block[self.index]
-        if self.current_block_shape == "Z_rev":
-            return self.get_current_block[self.index]
 
     def full_line_clear(self):
         y_coordinates = [y for (x, y) in self.landed_blocks]
--- a/tetris.py
+++ b/tetris.py
@@ -112,7 +112,7 @@ class TetrisGame:
         self.upcoming_block_shape = None
         self.current_block_shape = None
         self.current_block_center = None
-        new_block()
+
 
     def new_block(self):
         """
--- a/tetris.py
+++ b/tetris.py
@@ -112,6 +112,7 @@ class TetrisGame:
         self.upcoming_block_shape = None
         self.current_block_shape = None
         self.current_block_center = None
+        new_block()
 
     def new_block(self):
         """
@@ -122,7 +123,7 @@ class TetrisGame:
             self.current_block_shape = random.choice(shape_names)
         else:
             self.current_block_shape = self.upcoming_block_shape
-        self.current_block_center = (int(game_width / 2), -2)
+        self.current_block_center = (int(game_width / 2), 8)
         self.upcoming_block_shape = random.choice(shape_names)
         self.index = 0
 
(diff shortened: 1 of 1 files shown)


tetris.py
```diff
@@ -125,38 +124,47 @@ def new_block(self):
             self.current_block_shape = self.upcoming_block_shape
         self.current_block_center = (int(game_width / 2), -2)
         self.upcoming_block_shape = random.choice(shape_names)
+        self.index = 0
 
-    def get_current_block(self):
+    def get_current_block(self, index):
```
Akuli, line 129:
```suggestion
    def get_current_block(self):
```

The `index` argument is never actually used. The method uses `self.index` which comes from `self` argument , not from `index` argument. This also simplifies all code that calls this method.

tetris.py
```diff
@@ -125,38 +124,47 @@ def new_block(self):
...
         (x, y) = self.current_block_center
         if self.current_block_shape == "I":
-            return [(x - 2, y), (x - 1, y), (x, y), (x + 1, y)]
+            self.I = [[(x - 2, y), (x - 1, y), (x, y), (x + 1, y)], [(x, y-3), (x, y-2), (x, y-1), (x, y)]]
+            return self.I[self.index]
```
Akuli, line 133: Two things:
1. Because this variable is not used anywhere else, it should be a local variable, not something in `self`.
2. You are not using `%` like we discussed.

```suggestion
            I = [[(x - 2, y), (x - 1, y), (x, y), (x + 1, y)], [(x, y-3), (x, y-2), (x, y-1), (x, y)]]
            return I[self.index % len(I)]
```

Other parts of this function need a similar change.

tetris.py
```diff
@@ -178,7 +186,22 @@ def block_mover(self):
...
+        if self.current_block_shape == "L_rev":
+            return self.get_current_block[self.index]
+        if self.current_block_shape == "O":
+            return
+        if self.current_block_shape == "E":
+            return self.get_current_block[self.index]
+        if self.current_block_shape == "Z":
+            return self.get_current_block[self.index]
+        if self.current_block_shape == "Z_rev":
+            return self.get_current_block[self.index]
```
Akuli, line 204: I suggest deleting all this, since incrementing `index` is enough to make `get_current_block()` to do the right thing, and hence get the block to show up as rotated:

```suggestion
```

Akuli (changes requested):

tetris.py
```diff
@@ -166,19 +175,20 @@ def block_mover(self):
...
         else:
             x, y = self.current_block_center
-            self.current_block_center = (x, y + 1)
+            self.current_block_center = (x, y)
 
     def block_rotator(self, event):
-        print("TODO: should rotate")
+        self.index += 1
+        if self.index > 4:
+            self.index = 1
```
Akuli, line 191: These lines are not necessary because of the `%` usage in  `get_current_block`, and can be deleted safely:
```suggestion
```

tetris.py
```diff
@@ -123,25 +123,33 @@ def new_block(self):
...
 
     def get_current_block(self):
         (x, y) = self.current_block_center
         if self.current_block_shape == "I":
-            return [(x - 2, y), (x - 1, y), (x, y), (x + 1, y)]
+            I = [[(x - 2, y), (x - 1, y), (x, y), (x + 1, y)], [(x, y-3), (x, y-2), (x, y-1), (x, y)]]
+            return I[self.index % len(I)]
         if self.current_block_shape == "L":
-            return [(x - 1, y - 1), (x - 1, y), (x, y), (x + 1, y)]
+            L = [[(x-1, y), (x,y), (x+1, y), (x+1, y-1)], [(x,y), (x-1,y), (x-1,y-1), (x-1,y-2)], [(x, y), (x, y-1), (x+1, y-1), (x+2, y-1)], [(x, y-1), (x+1, y-1), (x+1, y), (x+1,y+1)]]
```
Akuli, line 136: If you find these long lines difficult to work with, you can black the code. It will be blacked before merging anyway.

Akuli (changes requested):

tetris.py
```diff
@@ -120,28 +120,60 @@ def new_block(self):
...
         self.upcoming_block_shape = random.choice(shape_names)
+        self.index = 0
 
     def get_current_block(self):
         (x, y) = self.current_block_center
         if self.current_block_shape == "I":
-            return [(x - 2, y), (x - 1, y), (x, y), (x + 1, y)]
+            I = [
+                [(x - 2, y), (x - 1, y), (x, y), (x + 1, y)],
+                [(x, y - 3), (x, y - 2), (x, y - 1), (x, y)],
```
Akuli, line 135: I guess it would make sense to keep the center `(x, y)` exactly as it is, and only change other coords:

```suggestion
                [(x, y - 2), (x, y - 1), (x, y), (x, y + 1)],
```

Please change other blocks similarly, so that if one block contains `(x, y)`, then they all contain `(x, y)` in the same place.

Akuli (changes requested):

tetris.py
Arrinao: blacked

tetris.py
```diff
@@ -125,23 +125,55 @@ def new_block(self):
             self.current_block_shape = self.upcoming_block_shape
         self.current_block_center = (int(game_width / 2), -2)
         self.upcoming_block_shape = random.choice(shape_names)
+        self.index = 0
```
Akuli, line 128:
```suggestion
        self.rotate_counter = 0
```

tetris.py
```diff
@@ -125,23 +125,55 @@ def new_block(self):
...
+                [(x, y - 2), (x, y - 1), (x, y), (x, y + 1)],
+            ]
+            return I[self.index % len(I)]
         if self.current_block_shape == "L":
-            return [(x - 1, y - 1), (x - 1, y), (x, y), (x + 1, y)]
+            L = [
+                [(x - 1, y), (x, y), (x + 1, y), (x + 1, y - 1)],
+                [(x + 1, y), (x, y), (x, y - 1), (x, y - 2)],
+                [(x, y), (x, y - 1), (x + 1, y - 1), (x + 2, y - 1)],
+                [(x, y - 1), (x + 1, y - 1), (x + 1, y), (x + 1, y + 1)],
```
Akuli, line 143: This code is difficult to read. The coordinates don't follow any kind of obvious pattern. For example, the first coordinates of each list are `(x-1, y), (x+1, y), (x,y), (x, y-1)`. I have no idea whether that's what they should be.

If you would simply rotate around `(x, y)`, going 90 degrees clockwise at a time, you would get `(x-1, y), (x, y-1), (x+1,y), (x, y+1)`. This pattern is easy to pick up when reading to code, and so it's possible to properly review this code and check that the list is correct.

tetris.py
```diff
@@ -125,23 +125,55 @@ def new_block(self):
...
+                [(x, y), (x, y - 1), (x + 1, y - 1), (x + 1, y - 2)],
+            ]
+            return Z[self.index % len(Z)]
         if self.current_block_shape == "Z_rev":
-            return [(x - 1, y), (x, y), (x + 1, y), (x + 1, y + 1)]
+            Z_rev = [
+                [(x + 1, y - 1), (x, y - 1), (x, y), (x - 1, y)],
+                [(x, y), (x, y - 1), (x - 1, y - 1), (x - 1, y - 2)],
+            ]
+            return Z_rev[self.index % len(Z_rev)]
```
Akuli, line 176: Cleanup idea: Every part of this function ends with `return some_variable[self.index % len(some_variable)]`. If you use the same variable name for every variable, let's say `coord_lists`, it becomes `return coord_lists[self.index % len(coord_lists)]`, and then because the exact same code is used in each case, it can be written only once, at the end of the function.

Akuli (changes requested):
It's starting to look quite good :D I left some ideas for cleaning up the code.

tetris.py
```diff
@@ -106,6 +109,10 @@ def right_mediator(self, event):
         self.tetris_game.user_input_right()
         self.draw_block()
 
+    def rotate_mediator(self, event):
+        self.tetris_game.block_rotator()
+        print('kurva')
```
Akuli, line 114:
```suggestion
```

tetris.py
```diff
@@ -177,8 +216,13 @@ def block_mover(self):
...
             self.current_block_center = (x, y + 1)
 
-    def block_rotator(self, event):
-        print("TODO: should rotate")
+    def block_rotator(self):
+        '''
+        Rotates the current block
+        '''
+        self.index += 1
+        return self.get_current_block()
```
Akuli, line 224: The return value is not used when `block_rotator()` method is called, so it is not necessary:

```suggestion
```

Akuli (changes requested):

tetris.py
```diff
@@ -125,29 +132,63 @@ def new_block(self):
...
         self.upcoming_block_shape = random.choice(shape_names)
+        self.rotate_counter = 0
 
     def get_current_block(self):
         (x, y) = self.current_block_center
         if self.current_block_shape == "I":
-            return [(x - 2, y), (x - 1, y), (x, y), (x + 1, y)]
+            I = [
+                [(x - 2, y), (x - 1, y), (x, y), (x + 1, y)],
+                [(x, y + 1), (x, y), (x, y - 1), (x, y - 2)],
```
Akuli, line 142: I would arrange the coordinates so that below each coordinate is a rotated version of that coordinate. For example, below `(x - 2, y)` is `(x, y - 2)`. That way it's easier to see what's going on.

```suggestion
                [(x - 2, y), (x - 1, y), (x, y), (x + 1, y)],
                [(x, y - 2), (x, y - 1), (x, y), (x, y + 1)],
```

tetris.py
```diff
@@ -125,29 +132,63 @@ def new_block(self):
...
+            ]
+            return I[self.rotate_counter % len(I)]
         if self.current_block_shape == "L":
-            return [(x - 1, y - 1), (x - 1, y), (x, y), (x + 1, y)]
+            L = [
+                [(x - 1, y + 1), (x, y + 1), (x + 1, y + 1), (x + 1, y)],
+                [(x - 1, y - 1), (x - 1, y), (x - 1, y + 1), (x, y + 1)],
+                [(x - 1, y - 1), (x, y - 1), (x + 1, y - 1), (x - 1, y)],
+                [(x, y - 1), (x + 1, y - 1), (x + 1, y), (x + 1, y + 1)],
```
Akuli, line 150: Same here, changing so that below each coordinate is a rotated version of that coordinate:

```suggestion
                [(x + 1, y - 1), (x, y - 1), (x - 1, y - 1), (x - 1, y)],
                [(x + 1, y + 1), (x + 1, y), (x + 1, y - 1), (x, y - 1)],
```

Akuli (changes requested):

tetris.py
Arrinao: Man, as soon as I get used to a system *you* have suggested, you come up with another one 🤪

PR #28, opened 2021-02-26
Fixed the issue with rotating near borders or another blocks

Arrinao:
I just hope we won't end up on really_new_borderline_rotate branch

### DIFF
--- a/tetris.py
+++ b/tetris.py
@@ -222,12 +222,8 @@ class TetrisGame:
         Rotates the current block
         """
         self.rotate_counter += 1
-        if any(x <= -1 for (x, y) in self.get_current_block()) or any(
-            (x - 1, y) in self.landed_blocks for x, y in self.get_current_block()
-        ):
-            self.rotate_counter -= 1
-        if any(x >= game_width for x, y in self.get_current_block()) or any(
-            (x + 1, y) in self.landed_blocks for x, y in self.get_current_block()
+        if any(x <= -1 or x >= game_width for (x, y) in self.get_current_block()) or any(
+            (x, y) in self.landed_blocks for x, y in self.get_current_block()
         ):
             self.rotate_counter -= 1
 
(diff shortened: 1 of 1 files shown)


tetris.py
```diff
@@ -222,8 +222,19 @@ def block_rotator(self):
         Rotates the current block
         """
         self.rotate_counter += 1
+        if any(x <= -1 for (x, y) in self.get_current_block()) or any(
+            (x - 1, y) in self.landed_blocks for x, y in self.get_current_block()
+        ):
+            self.rotate_counter -= 1
+        if any(x >= game_width for x, y in self.get_current_block()) or any(
+            (x + 1, y) in self.landed_blocks for x, y in self.get_current_block()
+        ):
```
Akuli, line 231:
```suggestion
        if any(x <= -1 or x >= game_width for (x, y) in self.get_current_block()) or any(
            (x, y) in self.landed_blocks for x, y in self.get_current_block()
        ):
```

Akuli (changes requested):

tetris.py
```diff
@@ -222,8 +222,15 @@ def block_rotator(self):
         Rotates the current block
         """
         self.rotate_counter += 1
+        if any(x <= -1 or x >= game_width for (x, y) in self.get_current_block()) or any(
+            (x, y) in self.landed_blocks for x, y in self.get_current_block()
+        ):
```
Akuli, line 227:
```suggestion
        if any(x not in range(game_width) or (x, y) in self.landed_blocks for (x, y) in self.get_current_block()):
```

Akuli (changes requested):

tetris.py
Arrinao: Would you mind me keeping the above code commented out?

The last one is nice but it's something I wouldn't come up with at the moment :)

Akuli: We can keep it as is. It already works.

Akuli (approved):
Now `black` and we can merge.

PR #31, opened 2021-03-03
Added a condition to check for the bottom border and prevent rotation from going off grid.

Arrinao:


tetris.py
```diff
@@ -226,7 +226,9 @@ def block_rotator(self):
         #    (x, y) in self.landed_blocks for x, y in self.get_current_block()
         # ):
         if any(
-            x not in range(game_width) or (x, y) in self.landed_blocks
+            x not in range(game_width)
+            or y not in range(game_height)
```
Akuli, line 230: When the block is not fully visible yet, `y` will be negative and therefore not in `range(game_height)`. It should still be possible to move the block:

```suggestion
            or y >= game_height
```

Akuli (changes requested):

PR #32, opened 2021-03-05
Experimental

Arrinao:
Changed dicts for lists with tuples. Functionality seems to be the same.

### DIFF
--- a/tetris.py
+++ b/tetris.py
@@ -83,8 +83,17 @@ class TetrisGUI:
 
         self.canvas.delete("block")
         for x, y in (
-            self.tetris_game.get_current_block() + list(self.tetris_game.coord_extractor())
+            self.tetris_game.get_current_block()
         ):
+            self.canvas.create_rectangle(
+                x * square_size,
+                y * square_size,
+                x * square_size + square_size,
+                y * square_size + square_size,
+                tags="block",
+                fill=RED
+            )
+        for x, y in list(self.tetris_game.coord_extractor()):
             self.canvas.create_rectangle(
                 x * square_size,
                 y * square_size,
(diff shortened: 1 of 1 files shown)


tetris.py
```diff
@@ -175,12 +192,34 @@ def get_current_block(self):
 
         return coords[self.rotate_counter % len(coords)]
 
+    def coord_extractor(self):
+        coords = []
+        for square in self.landed_blocks:
+            coords.append(square[1])
+        return coords
```
Akuli, line 199: If you write it like this, then the person reading the code doesn't have to guess what `square[1]` does:

```suggestion
        return [coords for shape, coords in self.landed_blocks]
```

tetris.py
```diff
@@ -80,15 +85,27 @@ def draw_block(self):
...
+        for shape_letter, (x, y) in self.tetris_game.landed_blocks:
+            print(shape_letter, (x, y))
+            color = color_dict[shape_letter]
+            self.canvas.create_rectangle(
+                x * square_size,
+                y * square_size,
+                x * square_size + square_size,
+                y * square_size + square_size,
+                tags="block",
+                fill=color
```
Akuli, line 108: The review comment from #19 applies here too: if this copy/pasta bothers you, you can move it into a separate method.

tetris.py
```diff
@@ -175,12 +192,34 @@ def get_current_block(self):
...
+
+    def block_mover(self):
+        """
+        Moves the current block downwards one square on the canvas
+        """
+        if any(
+            (x, y + 1) in self.coord_extractor() for (x, y) in self.get_current_block()
+        ) or any(y + 1 == game_height for (x, y) in self.get_current_block()):
+            for coord in self.get_current_block():
+#               self.landed_blocks[self.current_block_shape]= coord   ##Doesn't work, only one coord is created :(
```
Akuli, line 209:
```suggestion
```

tetris.py
```diff
@@ -175,12 +192,34 @@ def get_current_block(self):
 
         return coords[self.rotate_counter % len(coords)]
 
+    def coord_extractor(self):
```
Akuli, line 195:
```suggestion
    def get_landed_coords(self):
```

tetris.py
```diff
@@ -223,24 +247,23 @@ def block_rotator(self):
...
-        for y_line in range(game_height):
-            count = coordinates_counter[y_line]
+        for x_line in range(game_height):
+            count = coordinates_counter[x_line]
             if count == game_width:
                 # TODO: root.after() here
-                self.landed_blocks = [
-                    (a, b + 1) for (a, b) in self.landed_blocks if b < y_line
-                ] + [(a, b) for (a, b) in self.landed_blocks if b > y_line]
+                self.landed_blocks = [(letter, (a, b)) for (letter, (a, b)) in self.landed_blocks if b > x_line] + [(letter, (a, b+1)) for (letter, (a, b)) in self.landed_blocks if b < x_line]
```
Akuli, line 266: Needs blacking

Akuli (changes requested):

PR #33, opened 2021-03-06
Experimental3

Arrinao:
line 218, 264, 280

### DIFF
--- a/tetris.py
+++ b/tetris.py
@@ -7,13 +7,16 @@ square_size = 35
 game_width = 10
 game_height = 15
 BLACK = "#000000"
-BLUE = "#0029ff"
-RED = "#ff1700"
-GREEN = "#05ff00"
+BLUE = "RoyalBlue2"
+RED = "red2"
+GREEN = "lime green"
 GREY = "#666666"
 D_GREY = "#383838"
-YELLOW = "#ffd343"
-PURPLE = "7e1e9c"
+YELLOW = "gold"
+PURPLE = "#9900FF"
+ORANGE = 'salmon1'
+PINK = '#FF00FF'
+TEAL = 'pale turquoise'
 
 shape_names = ["I", "L", "L_rev", "O", "E", "Z", "Z_rev"]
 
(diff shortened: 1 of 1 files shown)


tetris.py
```diff
@@ -248,7 +285,7 @@ def block_rotator(self):
         if any(
             x not in range(game_width)
             or y >= game_height
-            or (x, y) in self.get_landed_coords()
+            or (x, y) in self.coord_extractor()  # This is so hard to conceive by myself is this an actual comprehension of some sort?
```
Akuli, line 288: This is of the form `any(long_condition for (x, y) in self.get_current_block())`. So we go through coords of current block, in a list comprehension (which isn't actually a list comprehension, since there's no `[` and `]`), and check whether the condition is true for any coords.

```suggestion
            or (x, y) in self.coord_extractor()
```

tetris.py
```diff
@@ -257,21 +294,17 @@ def full_line_clear(self):
...
-                    (letter, (a, b))
-                    for (letter, (a, b)) in self.landed_blocks
-                    if b > x_line
-                ] + [
-                    (letter, (a, b + 1))
-                    for (letter, (a, b)) in self.landed_blocks
-                    if b < x_line
-                ]
+                for letter, coord_list in self.landed_blocks.items():
+                    # self.landed_blocks = {letter: [(a, b) for (a, b) in coord_list if b > y_line] + [(a, b+1) for (a, b) in coord_list if b < y_line]} #Why this doesn't work?
```
Akuli, line 304:
```suggestion
```

The code in this comment does `self.landed_blocks = {letter: something}`. Note that there is no `for` involved, other than inside lists, so this is not a comprehension, just a dict. The created dict has one key, `letter`, with a list as a value. So it replaces `self.landed_blocks` with a new dict that is missing all keys except the one `letter` key.

Akuli (changes requested):

tetris.py
```diff
@@ -7,16 +7,16 @@
...
 GREY = "#666666"
-D_GREY = "#383838"
+D_GREY = "gray7"
 YELLOW = "gold"
 PURPLE = "#9900FF"
-ORANGE = "salmon1"
+ORANGE = "Orangered2"
 PINK = "#FF00FF"
-TEAL = "pale turquoise"
+TEAL = "paleturquoise3"
```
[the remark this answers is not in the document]
Arrinao: The new ones

tetris.py
Arrinao: So essentially the comprehension itself happens only inside the list, and not the whole dict.

Akuli (approved):

PR #34, opened 2021-03-09
Topbar added, only as a Label so far

Arrinao:


tetris.py
```diff
@@ -291,10 +294,10 @@ def full_line_clear(self):
             if count == game_width:
                 # TODO: root.after() here
                 for letter, coord_list in self.landed_blocks.items():
-                    # self.landed_blocks = {letter: [(a, b) for (a, b) in coord_list if b > y_line] + [(a, b+1) for (a, b) in coord_list if b < y_line]} #Why this doesn't work?
+                    # self.landed_blocks = {letter: [(a, b) for (a, b) in coord_list if b > x_line] + [(a, b+1) for (a, b) in coord_list if b < x_line]} #Why this doesn't work?
                     self.landed_blocks[letter] = [
-                        (a, b) for (a, b) in coord_list if b > y_line
-                    ] + [(a, b + 1) for (a, b) in coord_list if b < y_line]
+                        (a, b) for (a, b) in coord_list if b > x_line
+                    ] + [(a, b + 1) for (a, b) in coord_list if b < x_line]
```
Akuli, line 300: Is the code on main branch this broken? I'm a terrible reviewer.

tetris.py
```diff
@@ -27,9 +27,12 @@ def run_gui():
     root = tkinter.Tk()
     root.resizable(False, False)
 
-    topbar = tkinter.Label(root)
-    topbar['bg'] = D_GREY
+    topbar = tkinter.Frame(root, bg=D_GREY, height=square_size*2, relief = 'ridge', borderwidth=1)
     topbar.grid(row=0, column=0, sticky='we')
+
+    topbar_next_block = tkinter.Label(topbar)
```
Akuli, line 33: I guess this should be a `Canvas`, if you want to actually display a block, rather than just the shape name.

tetris.py
```diff
@@ -27,9 +27,12 @@ def run_gui():
...
+    topbar_next_block.grid(column=2, columnspan=2, sticky='w')
```
Akuli, line 34: Because there's nothing else in the topbar, the column and columnspan don't matter, and defaults are fine:

```suggestion
    topbar_next_block.grid(sticky='w')
```

Akuli (changes requested):

tetris.py
Akuli: Not anymore, I fixed it

tetris.py
Arrinao: Adjusted

tetris.py
Arrinao: Adjusted

tetris.py
```diff
@@ -27,18 +28,41 @@ def run_gui():
...
         highlightthickness=0,
+        borderwidth=1
     )
-    game_canvas.grid()
+    game_canvas.grid(row=1, sticky='nswe')
 
-    tetris_gui = TetrisGUI(game_speed, game_canvas)
+    topbar = tkinter.Frame(root, bg=D_GREY, height=square_size*2, relief = 'ridge')
+    topbar.grid(row=0, columnspan=2, sticky='we')
+    topbar.columnconfigure(0, weight=1)
```
Akuli, line 42: This is not necessary, because you are using `pack` inside topbar, not `grid`.

tetris.py
```diff
@@ -27,18 +28,41 @@ def run_gui():
...
+
+    tetris_gui = TetrisGUI(game_speed, game_canvas, topbar_next_block)
+
+    topbar_time = tkinter.Label(topbar, bg=D_GREY, text=tetris_gui.timer(), font = 'digital-7', fg='orange', borderwidth=1)
```
Akuli, line 46: `text=tetris_gui.timer()` doesn't work, because it has to update once per second or so

tetris.py
```diff
@@ -27,18 +28,41 @@ def run_gui():
...
         width=square_size * game_width,
         height=square_size * game_height,
         highlightthickness=0,
+        borderwidth=1
     )
-    game_canvas.grid()
+    game_canvas.grid(row=1, sticky='nswe')
 
-    tetris_gui = TetrisGUI(game_speed, game_canvas)
+    topbar = tkinter.Frame(root, bg=D_GREY, height=square_size*2, relief = 'ridge')
```
Akuli, line 40: I don't think you need to provide the size here, because inside the topbar, there's a canvas with the correct size anyway:

```suggestion
    topbar = tkinter.Frame(root, bg=D_GREY, relief = 'ridge')
```

tetris.py
```diff
@@ -27,18 +28,41 @@ def run_gui():
...
+    topbar_time = tkinter.Label(topbar, bg=D_GREY, text=tetris_gui.timer(), font = 'digital-7', fg='orange', borderwidth=1)
+    topbar_time.pack(side='left', padx=10)
+
+    topbar_score = tkinter.Label(topbar, bg=D_GREY, text='foo', font = 'digital-7', fg='orange', borderwidth=1)
+    topbar_score.pack(side='left', fill='x', expand=True)
+
+    topbar_next_block = tkinter.Canvas(topbar, bg=D_GREY, width = square_size*4, height = square_size*2)
+    topbar_next_block.pack(side='right')
+
+    sidebar = tkinter.Frame(root, bg=D_GREY, height=square_size*game_height)
```
Akuli, line 55: What happens if you get rid of height, like this?

```suggestion
    sidebar = tkinter.Frame(root, bg=D_GREY)
```

tetris.py
```diff
@@ -27,18 +28,41 @@ def run_gui():
...
+    sidebar.grid(row=1, column=1, sticky='nsw')
+
+    new_game_button = tkinter.Button(sidebar, text = 'start')
+    new_game_button.grid(sticky='n')
+
+    new_game_button2 = tkinter.Button(sidebar, text = 'start')
+    new_game_button2.grid(sticky='n')
+
+    new_game_button3 = tkinter.Button(sidebar, text = 'start')
+    new_game_button3.grid(sticky='n')
```
Akuli, line 65: I would use `.pack()` instead of `.grid(sticky='n')`, it's slightly simpler

Akuli (changes requested):

tetris.py
Arrinao: Yes, I've changed that.

tetris.py
Arrinao: I'll use canvas.after for this but there seems to be some problem, I'll report when I narrow it down

tetris.py
```diff
@@ -27,26 +28,50 @@ def run_gui():
...
+
+    tetris_gui = TetrisGUI(game_speed, game_board)
+
+    main_board = Board(game_board, GREY, game_width, game_height, random.choice(shape_names), (int(game_width / 2), -2), 0)
```
Akuli, line 64: Variable names are confusing here: `game_board` is not actually a board, but a canvas.

tetris.py
```diff
@@ -292,10 +316,132 @@ def full_line_clear(self):
...
                 # TODO: root.after() here
                 for letter, coord_list in self.landed_blocks.items():
-                    # self.landed_blocks = {letter: [(a, b) for (a, b) in coord_list if b > y_line] + [(a, b+1) for (a, b) in coord_list if b < y_line]} #Why this doesn't work?
+                    # self.landed_blocks = {letter: [(a, b) for (a, b) in coord_list if b > x_line] + [(a, b+1) for (a, b) in coord_list if b < x_line]} #Why this doesn't work?
                     self.landed_blocks[letter] = [
                         (a, b) for (a, b) in coord_list if b > x_line
                     ] + [(a, b + 1) for (a, b) in coord_list if b < x_line]
 
 
+class Board:
```
Akuli, line 325: I think it would make more sense to have this class before `TetrisGUI`.

tetris.py
```diff
@@ -292,10 +316,132 @@ def full_line_clear(self):
...
+    def __init__(self, canvas, outline_color, width, height, block_shape, block_center, rotate_counter):
```
Akuli, line 326:
```suggestion
    def __init__(self, canvas, outline_color, width, height, block_center):
```

tetris.py
```diff
@@ -292,10 +316,132 @@ def full_line_clear(self):
...
+        self.canvas = canvas
+        self.outline_color = outline_color
+        self.width = width
+        self.height = height
+        self.block_shape = block_shape
```
Akuli, line 331:
```suggestion
        self.block_shape = random.choice(shape_letters)
```

tetris.py
```diff
@@ -292,10 +316,132 @@ def full_line_clear(self):
...
+        self.block_center = block_center
+        self.rotate_counter = rotate_counter
```
Akuli, line 333:
```suggestion
        self.rotate_counter = 0
```

tetris.py
```diff
@@ -292,10 +316,132 @@ def full_line_clear(self):
...
+            self.canvas_small.create_rectangle(
+                    x * square_size,
+                    y * square_size,
+                    x * square_size + square_size,
+                    y * square_size + square_size,
+                    tags="block",
+                    fill=self.color_dict[self.tetris_game.upcoming_block_shape],
+            )
+
+    def get_block_shape(self, block_shape, block_center, rotate_counter):
```
Akuli, line 402: The purpose of the class was to avoid all the arguments.

```suggestion
    def get_block_shape(self):
```

tetris.py
```diff
@@ -292,10 +316,132 @@ def full_line_clear(self):
...
+        (x, y) = block_center
```
Akuli, line 403: And instead use stuff conveniently from `self`:

```suggestion
        (x, y) = self.block_center
```

tetris.py
```diff
@@ -292,10 +316,132 @@ def full_line_clear(self):
...
+                [(x - 1, y - 1), (x, y - 1), (x, y), (x + 1, y)],
+                [(x + 1, y - 1), (x + 1, y), (x, y), (x, y + 1)],
+            ]
+        if block_shape == "Z_rev":
+            coords = [
+                [(x + 1, y - 1), (x, y - 1), (x, y), (x - 1, y)],
+                [(x + 1, y + 1), (x + 1, y), (x, y), (x, y - 1)],
+            ]
+
+        return coords[rotate_counter % len(coords)]
```
Akuli, line 443:
```suggestion
        return coords[self.rotate_counter % len(coords)]
```

tetris.py
```diff
@@ -292,10 +316,132 @@ def full_line_clear(self):
...
+                self.canvas.create_rectangle(
+                    x * square_size,
+                    y * square_size,
+                    x * square_size + square_size,
+                    y * square_size + square_size,
+                    tags="block",
+                    fill=self.color_dict[letter],
+                )
+
+        for x, y in self.get_block_shape(self.tetris_game.upcoming_block_shape, self.tetris_game.upcoming_block_center, 0):
```
Akuli, line 392: Why the third for loop?

tetris.py
```diff
@@ -292,10 +316,132 @@ def full_line_clear(self):
...
+            self.canvas.create_rectangle(
+                x * square_size,
+                y * square_size,
+                x * square_size + square_size,
+                y * square_size + square_size,
+                tags="block",
+                fill=self.color_dict[self.tetris_game.current_block_shape],
+            )
+
+        for letter, coord_list in self.tetris_game.landed_blocks.items():
```
Akuli, line 381: Use the argument (see my checklist):
```suggestion
        for letter, coord_list in landed_blocks.items():
```

tetris.py
```diff
@@ -292,10 +316,132 @@ def full_line_clear(self):
...
+            "E": GREEN,
+            "L_rev": BLUE,
+            "Z": PURPLE,
+            "Z_rev": TEAL,
+            "O": ORANGE,
+        }
+        self.canvas.delete("block")
+        self.canvas_small.delete('block')
+
+        for x, y in self.get_block_shape(self.tetris_game.current_block_shape, self.tetris_game.current_block_center, self.tetris_game.rotate_counter):
```
Akuli, line 371:
```suggestion
        for x, y in self.get_block_shape():
```

Akuli (changes requested):

tetris.py
Arrinao: I've changed to be the same as topbar_board. So I should do it the opposite way and name topbar_board topbar_canvas instead?

tetris.py
```diff
@@ -59,82 +84,98 @@ def __init__(self, speed, canvas):
         self.canvas = canvas
         self.rect_size = 25
         self.tetris_game = TetrisGame()
```
Akuli, line 86: To do 

```
self.main_board.draw_block(self.tetris_game.whatever)
```

like we discussed, you need to change TetrisGui's `__init__` like this (can't do a suggestion for whatever reason):

```
def __init__(self, speed, main_board):
    self.speed = speed
    self.main_board = main_board
    ...
```

Akuli (changes requested):

tetris.py
Arrinao: Done!

tetris.py
```diff
@@ -27,64 +28,162 @@ def run_gui():
...
+    new_game_button2 = tkinter.Button(sidebar, text = 'start')
+    new_game_button2.grid(sticky='n')
+
+    new_game_button3 = tkinter.Button(sidebar, text = 'start')
+    new_game_button3.grid(sticky='n')
+
+    tetris_gui = TetrisGUI(game_speed, main_canvas)
+
+    main_board = Board(main_canvas, GREY, game_width, game_height, random.choice(shape_names), (int(game_width / 2), -2), 0)
+    topbar_board = Board(topbar_canvas, D_GREY, square_size*4, square_size*2, random.choice(shape_names), (2, 1), 0)
```
Akuli, line 65: Now `tetris_gui` needs to take `main_board` instead of `main_canvas`, so create it after the boards and then pass in the board:

```suggestion
    main_board = Board(main_canvas, GREY, game_width, game_height, random.choice(shape_names), (int(game_width / 2), -2), 0)
    topbar_board = Board(topbar_canvas, D_GREY, square_size*4, square_size*2, random.choice(shape_names), (2, 1), 0)
    tetris_gui = TetrisGUI(game_speed, main_board)
```

Akuli (changes requested):

Akuli:
@Arrinao  Pull before you continue working on this branch, I merged main into it

Akuli:
replaced by #34 and #36

PR #37, opened 2021-03-21
Code cleanup

Arrinao:
Why isn't the game_canvas displayed? The rectangles are clearly there, because everything else works, but apparently fill=D_GREY and outline=self.outline_color does not compute :(


tetris.py
```diff
@@ -370,4 +312,38 @@ def full_line_clear(self):
...
+        self.main_board = main_board
+        self.start_time = time.time()
+        self.move_block()
+
+    def move_block(self):
+        """
+        Has the responsibility to call current_block_mover() and draw_block() to
+        simulate the blocks moving downwards on the canvas
+        """
+        self.main_board.current_block_mover()
```
Akuli, line 329: If `current_block_mover()` returns `True`, we make a new block:

```suggestion
        if self.main_board.current_block_mover():
            self.new_block()
```

You will need to add `new_block()` method.

tetris.py
```diff
@@ -285,7 +228,6 @@ def current_block_mover(self):
                 )
             )
             self.full_line_clear()
```
Akuli, line 230: Change `current_block_mover()` to return `True` when a new block should be added:

```suggestion
            self.full_line_clear()
            return True   # new block needed
```

tetris.py
```diff
@@ -285,7 +228,6 @@ def current_block_mover(self):
...
-            self.new_block()
         else:
             x, y = self.current_block_center
             self.current_block_center = (x, y + 1)
```
Akuli, line 233: And return `False` when a new block should not be added:

```suggestion
            self.current_block_center = (x, y + 1)
            return False  # no new block
```

Akuli (changes requested):
Currently new blocks are not created. The `new_block()` method can't go to `class Board` because it needs both boards (and not just one board, which is what you have inside `class Board`, only the board known as `self`). However, `main_board.current_block_mover()` is the method that knows whether `new_block()` should run or not. Here's how you can to communicate that information to `TetrisGui.move_block()`, the method that calls `current_block_mover`.

tetris.py
```diff
@@ -260,7 +209,7 @@ def get_block_shape(self, block_shape, block_center, rotate_counter):
     def get_landed_coords(self):
         return [coords for shape, coords in self.landed_blocks]
 
-    def current_block_mover(self):
+    def current_block_mover(self, small_board):
```
Akuli, line 212:
```suggestion
    def current_block_mover(self):
```

tetris.py
```diff
@@ -69,46 +69,49 @@ def run_gui():
...
+        self.width = width
+        self.height = height
+        self.outline_color = outline_color
+        self.landed_blocks = {}
+        self.current_block_center = current_block_center
+        self.current_block_shape = random.choice(shape_letter)
+        self.rotate_counter = 0
+        self.draw_board()
+        self.small_board = small_board
+        self.current_block_mover(small_board)
```
Akuli, line 97:
```suggestion
        self.current_block_mover()
```

tetris.py
```diff
@@ -154,65 +158,19 @@ def draw_block(self):
...
-            self.current_block_shape = random.choice(shape_names)
-        else:
-            self.current_block_shape = self.upcoming_block_shape
         self.current_block_center = (int(game_width / 2), -2)
-        self.upcoming_block_shape = random.choice(shape_names)
-        self.upcoming_block_center = (2, 1)
+        self.block_letter = self.small_board.block_letter
+        self.small_board.block_letter = random.choice(block_letter)
+        self.draw_block()
+        print(self.small_board.block_letter)
```
Akuli, line 170:
```suggestion
```

tetris.py
```diff
@@ -154,65 +158,19 @@ def draw_block(self):
...
         self.rotate_counter = 0
+        self.get_block_shape(self.block_letter, self.current_block_center, self.rotate_counter)
+
 
     def get_block_shape(self, block_shape, block_center, rotate_counter):
```
Akuli, line 175: This method should use `self.block_shape` etc:

```suggestion
    def get_block_shape(self):
```

Akuli (changes requested):

tetris.py
```diff
@@ -20,7 +20,7 @@
 PINK = "#FF00FF"
 TEAL = "paleturquoise3"
 
-shape_names = ["I", "L", "L_rev", "O", "E", "Z", "Z_rev"]
+block_letter = ["I", "L", "L_rev", "O", "E", "Z", "Z_rev"]
```
Akuli, line 23: The name should end with `s`, as this list contains multiple block letter`s`:

```suggestion
block_letters = ["I", "L", "L_rev", "O", "E", "Z", "Z_rev"]
```

tetris.py
```diff
@@ -69,46 +69,51 @@ def run_gui():
...
-        self.start_time = time.time()
+        self.width = width
+        self.height = height
+        self.outline_color = outline_color
+        self.landed_blocks = {}
+        self.current_block_center = current_block_center
+        self.block_letter = random.choice(block_letter)
+        self.rotate_counter = 0
+        self.draw_board()
+        self.small_board = small_board
```
Akuli, line 98: It's a bit weird how `self.small_board = small_board` is between `draw_board` and `draw_block`. Maybe put it above `draw_board`?

tetris.py
```diff
@@ -154,108 +154,61 @@ def draw_block(self):
...
-        Has the responsibility to call current_block_mover() and draw_block() to
-        simulate the blocks moving downwards on the canvas
-        """
-        self.tetris_game.current_block_mover()
-        self.draw_block()
-        self.canvas.after(game_speed, self.move_block)
-
-    def move_block_left(self, event):
-        self.tetris_game.user_input_left()
+    def letter_generator(self):
```
Akuli, line 157: This function is not used. Delete it.

tetris.py
```diff
@@ -154,108 +154,61 @@ def draw_block(self):
...
-        else:
-            self.current_block_shape = self.upcoming_block_shape
         self.current_block_center = (int(game_width / 2), -2)
-        self.upcoming_block_shape = random.choice(shape_names)
-        self.upcoming_block_center = (2, 1)
+        self.block_letter = self.small_board.block_letter
+        self.small_board.block_letter = random.choice(block_letter)
+        self.small_board.draw_block()
         self.rotate_counter = 0
+        self.get_block_shape()
```
Akuli, line 167: Delete this line. It does nothing because `get_block_shape()` creates and returns a list, but the returned list is not used.

tetris.py
```diff
@@ -266,44 +219,43 @@ def current_block_mover(self):
...
+        self.canvas.after(game_speed, self.current_block_mover)
+
+#    def move_block(self):
+#        """
+#        Has the responsibility to call current_block_mover() and draw_block() to
+#        simulate the blocks moving downwards on the canvas
+#        """
+#        self.main_board.current_block_mover()
+#        self.main_board.draw_block()
+#        self.main_board.canvas.after(game_speed, self.move_block)
```
Akuli, line 247: Delete these commented lines

tetris.py
```diff
@@ -370,4 +316,28 @@ def full_line_clear(self):
                     ] + [(a, b + 1) for (a, b) in coord_list if b < x_line]
 
 
+
+class TetrisGUI:
+    def __init__(self, speed, main_board):
+        self.speed = speed
```
Akuli, line 322: Speed argument and `self.speed` attribute are never used. Delete them.

tetris.py
```diff
@@ -370,4 +316,28 @@ def full_line_clear(self):
...
+        self.rect_size = 25
```
Akuli, line 323: `self.rect_size` is not used either.

Akuli (changes requested):

tetris.py
Arrinao: I think not because in the get_block_shape we are performing a check 'if self_block_letter == 'I':'

it makes more sense to use a singular here.

tetris.py
Arrinao: gone

tetris.py
Arrinao: gone

tetris.py
Arrinao: Deleted, but I think we should reintroduce this in the future for Board class

tetris.py
Arrinao: gone

tetris.py
Akuli: The lines are still there!

tetris.py
Akuli: No, it's still there

tetris.py
Akuli: This hasn't gotten deleted either

Akuli (approved):

PR #39, opened 2021-03-26
Fixed timer function

Arrinao:
Timer function works but it's ugly

### DIFF
--- a/tetris.py
+++ b/tetris.py
@@ -39,7 +39,7 @@ def run_gui():
     topbar.grid(row=0, columnspan=2, sticky="we")
 
     topbar_time = tkinter.Label(
-        topbar, bg=D_GREY, text="time", font="digital-7", fg="orange", borderwidth=1
+        topbar, bg=D_GREY, text='00:00', font="digital-7", fg="orange", borderwidth=1
     )
     topbar_time.pack(side="left", padx=10)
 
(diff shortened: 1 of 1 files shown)


tetris.py
```diff
@@ -81,7 +81,7 @@ def run_gui():
 
     main_board.current_block_mover()
 
-    tetris_gui = TetrisGUI(main_board)
+    tetris_gui = TetrisGUI(main_board, topbar_time, game_canvas)
```
Akuli, line 84:
```suggestion
    tetris_gui = TetrisGUI(main_board, topbar_time)
```

tetris.py
```diff
@@ -295,9 +295,12 @@ def full_line_clear(self):
 
 
 class TetrisGUI:
-    def __init__(self, main_board):
+    def __init__(self, main_board, topbar_time, canvas):
```
Akuli, line 298:
```suggestion
    def __init__(self, main_board, topbar_time):
```

tetris.py
```diff
@@ -295,9 +295,12 @@ def full_line_clear(self):
...
         self.main_board = main_board
+        self.topbar_time = topbar_time
         self.start_time = time.time()
+        self.game_canvas = canvas
```
Akuli, line 302:
```suggestion
```

tetris.py
```diff
@@ -313,7 +316,7 @@ def rotate_block(self, event):
 
     def timer(self):
         game_time = time.time() - self.start_time
-        return f"{int(game_time / 60):02d}:{int(game_time % 60):02d}"
-
+        self.topbar_time.config(text = f"{int(game_time / 60):02d}:{int(game_time % 60):02d}")
+        self.game_canvas.after(1000, self.timer)
```
```suggestion
        self.topbar_time.after(1000, self.timer)
```

Akuli (changes requested):
Get rid of canvas  and use `topbar_time.after`. It is a widget, so it has `.after()` method

Akuli (approved):

PR #40, opened 2021-03-26
Added score function into Board class.

Arrinao:
A simple score function that adds 10 points per full_line_clear.

### DIFF
--- a/tetris.py
+++ b/tetris.py
@@ -115,7 +115,7 @@ class Board:
         self.topbar_score = topbar_score
         self.draw_board()
         self.draw_block()
-        self.initial_score = 0
+        self.game_score = 0
 
     def draw_board(self):
         """
@@ -298,8 +298,8 @@ class Board:
                 self.score()
 
     def score(self):
-        self.initial_score += 10
-        self.topbar_score.config(text=self.initial_score)
+        self.game_score += 10
+        self.topbar_score.config(text=self.game_score)
 
 
 class TetrisGUI:
@@ -307,7 +307,7 @@ class TetrisGUI:
         self.main_board = main_board
         self.topbar_time = topbar_time
         self.start_time = time.time()
-        self.initial_score = 0
+        self.game_score = 0
         self.timer()
 
     def move_block_left(self, event):
--- a/tetris.py
+++ b/tetris.py
@@ -57,6 +57,7 @@ def run_gui():
         width=square_size * 4,
         height=square_size * 2,
         highlightthickness=0,
+
     )
     topbar_canvas.pack(side="right", expand=True)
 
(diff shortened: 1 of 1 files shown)


tetris.py
```diff
@@ -69,14 +69,15 @@ def run_gui():
...
-    small_board = Board(topbar_canvas, 4, 2, D_GREY, (2, 1), None)
+    small_board = Board(topbar_canvas, 4, 2, D_GREY, (2, 1), None, None)
     main_board = Board(
         game_canvas,
         game_width,
         game_height,
         GREY,
         (int(game_width / 2), -2),
         small_board,
+        topbar_score
```
Akuli, line 80: Variable name is not great. How about `topbar_label`?

tetris.py
```diff
@@ -111,8 +112,10 @@ def __init__(self, canvas, width, height, outline_color, current_block_center, s
         self.block_letter = random.choice(block_letters)
         self.rotate_counter = 0
         self.small_board = small_board
+        self.topbar_score = topbar_score
         self.draw_board()
         self.draw_block()
+        self.initial_score = 0
```
Akuli, line 118: Why is this called `initial_score`? It is not always the initial score, it changes when the game is played.

tetris.py
```diff
@@ -292,13 +295,19 @@ def full_line_clear(self):
                     self.landed_blocks[letter] = [(a, b) for (a, b) in coord_list if b > x_line] + [
                         (a, b + 1) for (a, b) in coord_list if b < x_line
                     ]
+                self.score()
+
+    def score(self):
```
Akuli, line 300:
```suggestion
    def increment_score(self):
```

tetris.py
```diff
@@ -292,13 +295,19 @@ def full_line_clear(self):
...
+        self.initial_score += 10
+        self.topbar_score.config(text=self.initial_score)
 
 
 class TetrisGUI:
     def __init__(self, main_board, topbar_time):
         self.main_board = main_board
         self.topbar_time = topbar_time
         self.start_time = time.time()
+        self.initial_score = 0
```
Akuli, line 310: Delete this

Akuli (changes requested):

tetris.py
Arrinao: There are three labels, (actually one of them is canvas) that the topbar frame is divided into. topbar_time, topbar_score and topbar_canvas. Don't you think it's better to keep the name?

tetris.py
Arrinao: Changed to score

Akuli: doesn't seem like it? please push

Arrinao: Done

tetris.py
Arrinao: Deleted

Akuli:
There are merge conflicts, you need to pull on main and merge main into `score` branch

Akuli:
merged

Akuli (approved):
a bit weird to change flash colors in a score pr, but whatever :)

PR #41, opened 2021-03-27
Game over

### LINKED ISSUE #14: missing features
- [x] different colors for different blocks
- [x] deleting full lines
- [ ] game over detection
- [x] move block all the way down when arrow down is pressed
- [x] pause

Arrinao:
Trying to implement enum now, but I defined it in run_gui, so I guess I have to either move it to the class itself or make it global

Closes #14

### DIFF
--- a/tetris.py
+++ b/tetris.py
@@ -228,6 +228,7 @@ class Board:
             self.landed_blocks[self.block_letter].extend(self.get_block_shape())
             self.full_line_clear()
             self.new_block()
+            self.game_over()
         else:
             x, y = self.current_block_center
             self.current_block_center = (x, y + 1)
@@ -292,6 +293,13 @@ class Board:
                     self.landed_blocks[letter] = [(a, b) for (a, b) in coord_list if b > x_line] + [
                         (a, b + 1) for (a, b) in coord_list if b < x_line
                     ]
+
+    def game_over(self):
+        y_coordinates = [y for (x, y) in self.coord_extractor()]
+        print(y_coordinates)
+        print(game_height)
+        if any(y <= 0 for y in y_coordinates):
+            print('LOOOOOOOOOOSER!')
 
 
 class TetrisGUI:
--- a/tetris.py
+++ b/tetris.py
@@ -91,8 +91,6 @@ def run_gui():
     root.bind("<Left>", tetris_gui.move_block_left)
     root.bind("<Right>", tetris_gui.move_block_right)
     root.bind("<Up>", tetris_gui.rotate_block)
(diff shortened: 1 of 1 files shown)


tetris.py
```diff
@@ -293,6 +303,12 @@ def full_line_clear(self):
                         (a, b + 1) for (a, b) in coord_list if b < x_line
                     ]
 
+    def game_over(self):
```
Akuli, line 306: Please rename this to `check_game_over` or something. This function can be called even when the game is not actually over yet.

tetris.py
```diff
@@ -22,6 +23,8 @@
 
 block_letters = ["I", "L", "L_rev", "O", "E", "Z", "Z_rev"]
 
+game_status = Enum('game_status', 'in_progress, game_lost, game_won')
```
Akuli, line 26: Name this like a class, so that it will stand out, and won't be confused with local variables:

```suggestion
GameStatus = Enum('GameStatus', 'in_progress, game_lost, game_won')
```

tetris.py
```diff
@@ -113,6 +118,7 @@ def __init__(self, canvas, width, height, outline_color, current_block_center, s
         self.small_board = small_board
         self.draw_board()
         self.draw_block()
+        self.game_status = game_status
```
Akuli, line 121: No need to take it as argument:

```suggestion
        self.game_status = GameStatus.in_progress
```

Akuli (changes requested):

tetris.py
Arrinao: Renamed

tetris.py
Arrinao: Renamed to ENUM_game_status

tetris.py
Arrinao: Renamed to ENUM_game_status

tetris.py
Arrinao: Removed

tetris.py
```diff
@@ -22,6 +23,8 @@
 
 block_letters = ["I", "L", "L_rev", "O", "E", "Z", "Z_rev"]
 
+ENUM_game_status = Enum('game_status', 'in_progress, game_lost, game_won')
```
Akuli, line 26: just follow the fucking naming convention

```suggestion
GameStatus = Enum('GameStatus', 'in_progress, game_lost, game_won')
```

Akuli (changes requested):

Arrinao (commented):
It's good

PR #43, opened 2021-03-28
Added function for fast-moving block down. It works, but throws a lot of errors.

Arrinao:
Curiously enough, if I use self.canvas which is a correct syntax, the whole thing crashes.

### DIFF
--- a/tetris.py
+++ b/tetris.py
@@ -86,8 +86,8 @@ def run_gui():
     root.bind("<Left>", tetris_gui.move_block_left)
     root.bind("<Right>", tetris_gui.move_block_right)
     root.bind("<Up>", tetris_gui.rotate_block)
-    root.bind('<Down>', tetris_gui.move_block_down_press)
-    root.bind('<KeyRelease-Down>', tetris_gui.move_block_down_release)
+    root.bind("<Down>", tetris_gui.move_block_down_press)
+    root.bind("<KeyRelease-Down>", tetris_gui.move_block_down_release)
 
     root.title("Tetris – by The Philgrim, Arrinao, and Master Akuli")
     # root.iconphoto(False, tkinter.PhotoImage(file=image_name.png")) TODO: INSERT LATER
--- a/tetris.py
+++ b/tetris.py
@@ -86,8 +86,8 @@ def run_gui():
     root.bind("<Left>", tetris_gui.move_block_left)
     root.bind("<Right>", tetris_gui.move_block_right)
     root.bind("<Up>", tetris_gui.rotate_block)
-    root.bind('<Down>', tetris_gui.move_block_down)
-    root.bind('<KeyRelease-Down>', tetris_gui.move_block_down)
+    root.bind('<Down>', tetris_gui.move_block_down_press)
+    root.bind('<KeyRelease-Down>', tetris_gui.move_block_down_release)
 
     root.title("Tetris – by The Philgrim, Arrinao, and Master Akuli")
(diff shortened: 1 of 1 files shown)


tetris.py
```diff
@@ -309,6 +322,14 @@ def move_block_right(self, event):
         self.main_board.user_input_right()
         self.main_board.draw_block()
 
+    def move_block_down(self, event):
+        if self.main_board.moving_down is False:
+            self.main_board.moving_down = True
+        else:
+            self.main_board.moving_down = False
```
Akuli, line 329:
```suggestion
        if not self.main_board.moving_down:
            return
```

tetris.py
```diff
@@ -113,6 +115,7 @@ def __init__(self, canvas, width, height, outline_color, current_block_center, s
         self.small_board = small_board
         self.draw_board()
         self.draw_block()
+        self.moving_down = False
```
Akuli, line 118: Why is this in `Board`?

tetris.py
```diff
@@ -256,6 +259,16 @@ def user_input_right(self):
         x, y = self.current_block_center
         self.current_block_center = (x + 1, y)
 
+    def user_input_down(self):
+        if any((x, y + 1) in self.coord_extractor() for (x, y) in self.get_block_shape()) or any(
+            y + 1 == game_height for (x, y) in self.get_block_shape()
+        ):
+            return
+        while self.moving_down is True:
```
Akuli, line 267: `while` loop not necessary, just use `.after()`

Akuli (changes requested):

tetris.py
Arrinao: It's a flag that determines whether the loop in user_input_down should be active or not. Functions move_block_down_press and move_block_down_release affect its state

tetris.py
Arrinao: Done differently, check out the latest push

tetris.py
```diff
@@ -229,8 +232,9 @@ def current_block_mover(self):
             self.full_line_clear()
             self.new_block()
         else:
-            x, y = self.current_block_center
-            self.current_block_center = (x, y + 1)
+            if self.moving_down is False:
```
Akuli, line 235:
```suggestion
            if not self.moving_down:
```

tetris.py
```diff
@@ -256,6 +260,17 @@ def user_input_right(self):
         x, y = self.current_block_center
         self.current_block_center = (x + 1, y)
 
+    def user_input_down(self):
+        if any((x, y + 1) in self.coord_extractor() for (x, y) in self.get_block_shape()) or any(
+            y + 1 == game_height for (x, y) in self.get_block_shape()
+        ):
+            return
+        if self.moving_down is True:
```
Akuli, line 268:
```suggestion
        if self.moving_down:
```

tetris.py
```diff
@@ -309,6 +324,15 @@ def move_block_right(self, event):
         self.main_board.user_input_right()
         self.main_board.draw_block()
 
+    def move_block_down_press(self, event):
+        if self.main_board.moving_down is True:
```
Akuli, line 328:
```suggestion
        if self.main_board.moving_down:
```

Akuli (changes requested):
Code looks good overall and works well!

When checking for booleans, write `if thing` instead of `if thing is True`, just like you would write `if a == b` instead of `if (a == b) is True`. Also, use `not` instead of `is False`.

PR #44, opened 2021-03-29
Pause implemented, but doesn't seem to register the 'P' key

Arrinao:
I don't know if I need to create another function in Board class to have it perform stuff. But the self.paused state doesn't seem to change.

### DIFF
--- a/tetris.py
+++ b/tetris.py
@@ -86,6 +86,7 @@ def run_gui():
     root.bind("<Left>", tetris_gui.move_block_left)
     root.bind("<Right>", tetris_gui.move_block_right)
     root.bind("<Up>", tetris_gui.rotate_block)
+    root.bind('<P>', tetris_gui.pause_game)
 
     root.title("Tetris – by The Philgrim, Arrinao, and Master Akuli")
     # root.iconphoto(False, tkinter.PhotoImage(file=image_name.png")) TODO: INSERT LATER
@@ -113,6 +114,7 @@ class Board:
         self.small_board = small_board
         self.draw_board()
         self.draw_block()
+        self.paused = False
 
     def draw_board(self):
         """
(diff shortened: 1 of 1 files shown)


tetris.py
```diff
@@ -86,6 +86,7 @@ def run_gui():
     root.bind("<Left>", tetris_gui.move_block_left)
     root.bind("<Right>", tetris_gui.move_block_right)
     root.bind("<Up>", tetris_gui.rotate_block)
+    root.bind('<P>', tetris_gui.pause_game)
```
Akuli, line 89: you probably want to trigger when p is pressed, not shift+p:

```suggestion
    root.bind('<p>', tetris_gui.pause_game)
```

Akuli (changes requested):

tetris.py
```diff
@@ -314,13 +317,23 @@ def __init__(self, main_board, topbar_time):
         self.start_time = time.time()
         self.timer()
 
+    def pause_game(self, event):
+        if self.main_board.paused is True:
```
Akuli, line 321:
```suggestion
        if self.main_board.paused:
```

tetris.py
```diff
@@ -337,13 +350,14 @@ def move_block_down_release(self, event):
...
 
     def timer(self):
-        game_time = time.time() - self.start_time
-        self.topbar_time.config(text=f"{int(game_time / 60):02d}:{int(game_time % 60):02d}")
-        self.topbar_time.after(1000, self.timer)
-
+        if not self.main_board.paused:
+            game_time = time.time() - self.start_time
+            self.topbar_time.config(text=f"{int(game_time / 60):02d}:{int(game_time % 60):02d}")
+            self.topbar_time.after(1000, self.timer)
```
Akuli, line 361: Does the timer start running again when you unpause?

tetris.py
```diff
@@ -312,20 +313,32 @@ def __init__(self, main_board, topbar_time):
         self.main_board = main_board
         self.topbar_time = topbar_time
         self.start_time = time.time()
+        self.paused = False
         self.timer()
 
+    def pause_game(self, event):
+        if self.paused is True:
```
```suggestion
        if self.paused:
```

tetris.py
```diff
@@ -312,20 +313,32 @@ def __init__(self, main_board, topbar_time):
...
+            self.paused = False
+            self.move_block_down()
+        else:
+            if not self.paused:
+                self.paused = True
```
Akuli, line 325: No need to check anything, because the `else` never runs with `self.paused` set to `True`:
```suggestion
            self.paused = True
```

Akuli (changes requested):

tetris.py
```diff
@@ -312,20 +313,36 @@ def __init__(self, main_board, topbar_time):
...
         self.start_time = time.time()
+        self.pause_start = 0
+        self.paused_time = 0
+        self.paused = False
         self.timer()
 
+    def pause_game(self, event):
+        if self.paused:
+            self.paused = False
+            self.paused_time = time.time() - self.pause_start
```
Akuli, line 324:
```suggestion
            self.paused_time += time.time() - self.pause_start
```

Akuli (changes requested):

Akuli (approved):
Black and we can merge

PR #45, opened 2021-03-29
current block down moving .after() logic to TetrisGui

Akuli:
Will help with pause feature


Akuli:
Now pull on main, merge main into your branches, and fix conflicts.

PR #47, opened 2021-04-01
Added messy code for flashing the whole landed_blocks when a line is clear

Arrinao:
but it doesn't work

### DIFF
--- a/tetris.py
+++ b/tetris.py
@@ -138,6 +138,16 @@ class Board:
 
             x_gap += square_size
 
+    def draw_rectangle(self, x, y, tags, fill):
+        self.canvas.create_rectangle(
+                x * square_size,
+                y * square_size,
+                x * square_size + square_size,
+                y * square_size + square_size,
+                tags = tags,
+                fill = fill,
+        )
+
     def draw_block(self):
         """
         Draws the different shapes on the board
@@ -154,25 +164,11 @@ class Board:
         self.canvas.delete("block")
 
         for x, y in self.get_block_shape():
-            self.canvas.create_rectangle(
-                x * square_size,
-                y * square_size,
-                x * square_size + square_size,
-                y * square_size + square_size,
-                tags="block",
-                fill=self.color_dict[self.block_letter],
-            )
+            self.draw_rectangle(x, y, 'block', self.color_dict[self.block_letter])
 
         for letter, coord_list in self.landed_blocks.items():
             for (x, y) in coord_list:
-                self.canvas.create_rectangle(
-                    x * square_size,
-                    y * square_size,
-                    x * square_size + square_size,
-                    y * square_size + square_size,
-                    tags="block",
-                    fill=self.color_dict[letter],
-                )
+                self.draw_rectangle(x, y, 'block', self.color_dict[letter])
 
     def new_block(self):
         self.current_block_center = (int(game_width / 2), -2)
@@ -327,15 +323,7 @@ class Board:
         '''
         for x in range(game_width):
             for x_line in full_lines:
-                self.canvas.create_rectangle(
-                    x * square_size,
-                    x_line * square_size,
-                    x * square_size + square_size,
-                    x_line * square_size + square_size,
-                    tags="flash",
-                    fill=fill,
-                    outline=self.outline_color
-                )
+                self.draw_rectangle(x, x_line, 'flash', fill)
 
 
 class TetrisGUI:
--- a/tetris.py
+++ b/tetris.py
@@ -309,7 +309,6 @@ class Board:
                     self.flasher(x_line, 'brown')
                     self.canvas.update()
                     time.sleep(0.1)
-                # TODO: root.after() here
                 for letter, coord_list in self.landed_blocks.items():
                     # self.landed_blocks = {letter: [(a, b) for (a, b) in coord_list if b > x_line] + [(a, b+1) for (a, b) in coord_list if b < x_line]} #Why this doesn't work?
                     self.landed_blocks[letter] = [(a, b) for (a, b) in coord_list if b > x_line] + [
--- a/tetris.py
+++ b/tetris.py
@@ -304,17 +304,16 @@ class Board:
             count = coordinates_counter[x_line]
             if count == game_width:
                 flash_list.append(x_line)
-        for x_line in flash_list:
-            count = coordinates_counter[x_line]
-            if count == game_width:
-                for flash in range(2):
-                    self.flasher(flash_list, 'yellow')
-                    self.canvas.update()
-                    time.sleep(0.1)
-                    self.flasher(flash_list, 'blue')
-                    self.canvas.update()
-                    time.sleep(0.1)
+        if x_line in flash_list:
+            for flash in range(2):
+                self.flasher(flash_list, 'yellow')
+                self.canvas.update()
+                time.sleep(0.1)
+                self.flasher(flash_list, 'blue')
+                self.canvas.update()
+                time.sleep(0.1)
             self.canvas.delete('flash')
+        for x_line in flash_list:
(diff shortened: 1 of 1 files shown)


tetris.py
```diff
@@ -300,13 +301,45 @@ def full_line_clear(self):
         for x_line in range(game_height):
             count = coordinates_counter[x_line]
             if count == game_width:
+                print(coordinates_counter)
+                print(y_coordinates)
+                self.flasher()
```
Akuli, line 306: When this code runs, you have already figured out which line needs to be flashed, and there's no need to do it again:

```suggestion
                self.flasher(x_line)
```

tetris.py
```diff
@@ -300,13 +301,45 @@ def full_line_clear(self):
...
+                self.canvas.after(600)
```
Akuli, line 307: Don't use `.after()` with one argument. This is same as `time.sleep(0.6)`.

tetris.py
```diff
@@ -300,13 +301,45 @@ def full_line_clear(self):
...
+                    for x, y in self.get_block_shape():
+                        self.canvas.create_rectangle(
+                            x * square_size,
+                            y * square_size,
+                            x * square_size + square_size,
+                            y * square_size + square_size,
+                            tags="block",
+                            fill='White',
+                        )
+                    self.canvas.after(100)
```
Akuli, line 332: Same here, this is `time.sleep(0.1)`

Akuli (changes requested):

tetris.py
```diff
@@ -300,13 +302,31 @@ def full_line_clear(self):
...
             count = coordinates_counter[x_line]
             if count == game_width:
+                for flash in range(2):
+                    self.flasher(x_line, 'yellow')
+                    self.canvas.update()
+                    time.sleep(0.1)
+                    self.flasher(x_line, 'brown')
+                    self.canvas.update()
+                    time.sleep(0.1)
                 # TODO: root.after() here
```
Akuli, line 312:
```suggestion
```

tetris.py
```diff
@@ -300,13 +302,31 @@ def full_line_clear(self):
...
                 for letter, coord_list in self.landed_blocks.items():
                     # self.landed_blocks = {letter: [(a, b) for (a, b) in coord_list if b > x_line] + [(a, b+1) for (a, b) in coord_list if b < x_line]} #Why this doesn't work?
                     self.landed_blocks[letter] = [(a, b) for (a, b) in coord_list if b > x_line] + [
                         (a, b + 1) for (a, b) in coord_list if b < x_line
                     ]
 
+    def flasher(self, x_line, fill):
+        print('flash!')
+        for x in range(10):
```
Akuli, line 321:
```suggestion
        for x in range(game_width):
```

tetris.py
```diff
@@ -300,13 +302,31 @@ def full_line_clear(self):
...
+                    time.sleep(0.1)
                 # TODO: root.after() here
                 for letter, coord_list in self.landed_blocks.items():
                     # self.landed_blocks = {letter: [(a, b) for (a, b) in coord_list if b > x_line] + [(a, b+1) for (a, b) in coord_list if b < x_line]} #Why this doesn't work?
                     self.landed_blocks[letter] = [(a, b) for (a, b) in coord_list if b > x_line] + [
                         (a, b + 1) for (a, b) in coord_list if b < x_line
                     ]
 
+    def flasher(self, x_line, fill):
+        print('flash!')
```
```suggestion
```

Akuli (changes requested):

tetris.py
```diff
@@ -21,7 +21,7 @@
 TEAL = "paleturquoise3"
 WHITE = 'snow'
 
-block_letters = ["I", "L", "L_rev", "O", "E", "Z", "Z_rev"]
+block_letters = ["I"]
```
Akuli, line 24: Did you `git diff --cached` before commit? :)

tetris.py
```diff
@@ -297,35 +297,42 @@ def full_line_clear(self):
...
-                    self.flasher(x_line, 'brown')
-                    self.canvas.update()
-                    time.sleep(0.1)
-                self.flasher(x_line, D_GREY)
-                for letter, coord_list in self.landed_blocks.items():
-                    # self.landed_blocks = {letter: [(a, b) for (a, b) in coord_list if b > x_line] + [(a, b+1) for (a, b) in coord_list if b < x_line]} #Why this doesn't work?
-                    self.landed_blocks[letter] = [(a, b) for (a, b) in coord_list if b > x_line] + [
-                        (a, b + 1) for (a, b) in coord_list if b < x_line
+                flash_list.append(x_line)
+        if x_line in flash_list:
```
Akuli, line 307: This check is unnecessary and doesn't do what you think it does. Delete it.

tetris.py
```diff
@@ -297,35 +297,42 @@ def full_line_clear(self):
         """
         Clears the line once it's fully populated with blocks
         """
+        flash_list = []
```
Akuli, line 300: Maybe name this `full_lines`? It's not only for flashing, it's also used for deleting lines.

tetris.py
```diff
@@ -297,35 +297,42 @@ def full_line_clear(self):
...
+                self.canvas.update()
+                time.sleep(0.1)
+            self.canvas.delete('flash')
+        for x_line in flash_list:
+            for letter, coord_list in self.landed_blocks.items():
+                # self.landed_blocks = {letter: [(a, b) for (a, b) in coord_list if b > x_line] + [(a, b+1) for (a, b) in coord_list if b < x_line]} #Why this doesn't work?
+                self.landed_blocks[letter] = [(a, b) for (a, b) in coord_list if b > x_line] + [
+                    (a, b + 1) for (a, b) in coord_list if b < x_line
                     ]
+        print(flash_list)
```
Akuli, line 322: Delete this

tetris.py
```diff
@@ -297,35 +297,42 @@ def full_line_clear(self):
...
-                x_line * square_size + square_size,
-                fill=fill,
-                outline=self.outline_color
-            )
+            for x_line in flash_list:
+                self.canvas.create_rectangle(
+                    x * square_size,
+                    x_line * square_size,
+                    x * square_size + square_size,
+                    x_line * square_size + square_size,
```
Akuli, line 331: These calculations are copy/pasted to many places. Can we make a new method that does them?

tetris.py
Arrinao: No, but this is not a big deal. Will do it next time though :)

tetris.py
Arrinao: Changed to if flash_list:

tetris.py
Arrinao: Gone

tetris.py
Arrinao: Sure

tetris.py
```diff
@@ -293,20 +295,48 @@ def block_rotator(self):
...
+        Used in conjuction with full_line clear for flashing purposes
+        '''
+        for x in range(game_width):
+            for x_line in full_lines:
+                self.canvas.create_rectangle(
+                    x * square_size,
+                    x_line * square_size,
+                    x * square_size + square_size,
+                    x_line * square_size + square_size,
+                    tags="flash",
```
Akuli, line 335: The rest of the program doesn't use `flash` tag anywhere, but there is `self.canvas.delete("block")`:

```suggestion
                    tags="block",
```

Akuli (changes requested):

tetris.py
```diff
@@ -19,6 +19,7 @@
 ORANGE = "Orangered2"
 PINK = "#FF00FF"
 TEAL = "paleturquoise3"
+WHITE = "snow"
```
Akuli, line 22: delete this

Akuli (changes requested):

PR #48, opened 2021-04-05
"Can move" cleanup

Akuli:


Akuli:
Too difficult for Arrinao to understand, not worth the effort

PR #52, opened 2021-04-08
Added a canvas.after_cancel feature 

Arrinao:
to prevent speeding up the game due to multiple move_block_down calls when pausing the game


tetris.py
```diff
@@ -337,21 +337,29 @@ def __init__(self, main_board, topbar_time):
         self.paused_time = 0
         self.game_status = GameStatus.in_progress
         self.timer()
+        self.no_retry_pause = False
```
Akuli, line 340: The logic seems to be confusingly backwards here:

```suggestion
        retry_pause = True
```

Akuli:
Was fixed on main in ee0da88

PR #54, opened 2021-04-13
topbar_canvas now follows the shape of over new block EXCEPT for the …

Arrinao:
…first one. Also made an ugly hack for the I shape in order to fit it into the 1x4 canvas

### DIFF
--- a/tetris.py
+++ b/tetris.py
@@ -51,18 +51,20 @@ def run_gui():
     )
     topbar_score.pack(side="left", fill="x", expand=True)
 
-    topbar_holder = tkinter.Frame(topbar, bg=D_GREY, relief='ridge', height=square_size*3, width=square_size*5)
-    topbar_holder.pack(side='right', expand=True)
-    topbar_holder.pack_propagate(0)
+    topbar_canvas_container = tkinter.Frame(topbar, bg=D_GREY, relief='ridge', height=square_size*3, width=square_size*5)
+    topbar_canvas_container.pack(side='right', expand=True)
+    topbar_canvas_container.pack_propagate(0)
+
+    #canvas_container holds topbar_canvas. This is required to prevent stretching width of the topbar_canvas as its width changes when displaying certain shapes.
 
     topbar_canvas = tkinter.Canvas(
-        topbar_holder,
+        topbar_canvas_container,
         bg=D_GREY,
         width=square_size * 4,
         height=square_size * 2,
         highlightthickness=0,
     )
-    topbar_canvas.pack(side="right", expand=True, pady=15)
+    topbar_canvas.pack(side="right", expand=True)
 
     sidebar = tkinter.Frame(root, bg=D_GREY)
     sidebar.grid(row=1, column=1, sticky="nsw")
--- a/tetris.py
+++ b/tetris.py
@@ -38,7 +38,7 @@ def run_gui():
     )
     game_canvas.grid(row=1, sticky="nswe")
 
-    topbar = tkinter.Frame(root, bg=D_GREY, relief="ridge")
+    topbar = tkinter.Frame(root, bg=D_GREY, relief="ridge", height=square_size*50)
     topbar.grid(row=0, columnspan=2, sticky="we")
 
     topbar_time = tkinter.Label(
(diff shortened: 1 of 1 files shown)


tetris.py
```diff
@@ -51,14 +51,18 @@ def run_gui():
     )
     topbar_score.pack(side="left", fill="x", expand=True)
 
+    topbar_holder = tkinter.Frame(topbar, bg=D_GREY, relief='ridge', height=square_size*3, width=square_size*5)
```
Akuli, line 54: Please add a comment explaining why the canvas needs to go inside a separate frame.

Also stop naming things confusingly! This is not a "topbar holder", as it contains a canvas, instead of holding a topbar, whatever that would mean. How about `canvas_container`?

tetris.py
```diff
@@ -51,14 +51,18 @@ def run_gui():
...
+    topbar_holder.pack(side='right', expand=True)
+    topbar_holder.pack_propagate(0)
```
Akuli, line 56: What does this do? Is it necessary?

tetris.py
```diff
@@ -51,14 +51,18 @@ def run_gui():
...
     topbar_canvas = tkinter.Canvas(
-        topbar,
+        topbar_holder,
         bg=D_GREY,
         width=square_size * 4,
         height=square_size * 2,
         highlightthickness=0,
     )
-    topbar_canvas.pack(side="right", expand=True)
+    topbar_canvas.pack(side="right", expand=True, pady=15)
```
Akuli, line 65: Does this need the padding?

tetris.py
```diff
@@ -181,14 +196,38 @@ def new_block(self):
...
+            self.canvas.pack(pady=5)
+        if self.block_letter == 'Z_rev':
+            self.canvas.config(width=square_size * 3, height=square_size * 2)
+            self.canvas.pack(pady=5)
+
     def get_block_shape(self):
         (x, y) = self.current_block_center
         if self.block_letter == "I":
-            coords = [[(x - 2, y), (x - 1, y), (x, y), (x + 1, y)]]
+            coords = [[(x - 1, y - 1), (x, y - 1), (x + 1, y - 1), (x + 2, y - 1)]]
```
Akuli, line 230: This makes the `I` block rotate weirdly, because the center of rotation `(x, y)` is no longer on the block. Is this really needed? Can you solve the problem that this solves in `resize_to_fit` function maybe?

Akuli (changes requested):

tetris.py
Arrinao: Changed to topbar_canvas_container

tetris.py
Arrinao: Added a comment explaining its role.

tetris.py
Arrinao: Hmm. The padding actually doesn't seem to do anything....

tetris.py
```diff
@@ -51,8 +51,14 @@ def run_gui():
     )
     topbar_score.pack(side="left", fill="x", expand=True)
 
+    topbar_canvas_container = tkinter.Frame(topbar, bg=D_GREY, relief='ridge', height=square_size*3, width=square_size*5)
+    topbar_canvas_container.pack(side='right', expand=True)
+    topbar_canvas_container.pack_propagate(0)
+
+    #canvas_container holds topbar_canvas. This is required to prevent stretching width of the topbar_canvas as its width changes when displaying certain shapes.
```
Akuli, line 58: Split long comments to multiple lines so that they fit on the screen at once:

```suggestion
    # canvas_container holds topbar_canvas. This is required to prevent stretching
    # width of the topbar_canvas as its width changes when displaying certain shapes.
```

Also, please put this comment before `topbar_canvas_container = ...` so that I see it before I ask myself "wtf is `topbar_canvas_container`.

Akuli (changes requested):

tetris.py
Akuli: I think `resize_to_fit` sets its own padding after this code runs

tetris.py
Akuli: I was referring to `pack_propagate(0)`. [The manual](https://tcl.tk/man/tcl8.6/TkCmd/pack.htm) says that it disables propagation and "the  packer  will not set the requested width and height of the packer", but I don't really get it.

Arrinao: pack_propagate is by default set to 1 and makes the frame stretch in order to accomodate its windows. Setting it to 0 means height= and width= will stop being overlooked

tetris.py
Arrinao: You're right. I deleted it.

tetris.py
Arrinao: Message changed and moved

tetris.py
Akuli: Then maybe:

```
topbar_holder.pack_propagate(0)   # don't overlook width and height
```

tetris.py
```diff
@@ -181,14 +197,41 @@ def new_block(self):
...
         self.fast_down = False
 
+    def resize_to_fit(self):
+        if self.block_letter == 'I':
+            self.canvas.config(width=square_size * 4, height=square_size * 1)
+            self.canvas.pack(pady=square_size/2+5)
+            self.current_block_center = (1, 0)
+        else:
+            self.current_block_center = (1, 1)
+        if self.block_letter == 'L':
```
Akuli, line 212: A blank line above `if self.block_letter == 'L'` would help separate the first `if,else` from the rest

tetris.py
```diff
@@ -181,14 +197,41 @@ def new_block(self):
...
+            self.canvas.pack(pady=5)
+        if self.block_letter == 'Z_rev':
+            self.canvas.config(width=square_size * 3, height=square_size * 2)
+            self.canvas.pack(pady=5)
+
     def get_block_shape(self):
         (x, y) = self.current_block_center
         if self.block_letter == "I":
-            coords = [[(x - 2, y), (x - 1, y), (x, y), (x + 1, y)]]
+            coords = [[(x - 1, y), (x, y), (x + 1, y), (x + 2, y)]]
```
Akuli, line 234: Do we need to change these?

Arrinao: I suppose not anymore. I don't even remember why I did this

tetris.py
Arrinao: Done

Akuli (approved):

PR #56, opened 2021-04-15
Moved the self.draw_block call from __init__ to resize_to_fit. Should fix #55

### LINKED ISSUE #55: 'I' block is not displayed in small_board
We brought a bug to main:(

'I' block is not displayed in small_board if it's the first next_block (i.e. if resize_to_fit is not called from inside the class).

Arrinao:
fix #55

### DIFF
--- a/tetris.py
+++ b/tetris.py
@@ -21,7 +21,7 @@ ORANGE = "Orangered2"
 PINK = "#FF00FF"
 TEAL = "paleturquoise3"
 
-block_letters = ["I"]
+block_letters = ["I", "L", "L_rev", "O", "E", "Z", "Z_rev"]
 
 GameStatus = Enum("GameStatus", "in_progress, game_lost, paused")
 
@@ -89,9 +89,7 @@ def run_gui():
         None,
     )
     small_board.resize_to_fit()
-
-    print(small_board.block_letter)
-    print(small_board.current_block_center)
+    small_board.draw_block()
 
     main_board = Board(
         game_canvas,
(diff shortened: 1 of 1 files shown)


Arrinao:
How come 'fix #55' in the description doesn't register?

Akuli:
You put it to the title, not to the description

tetris.py
```diff
@@ -21,7 +21,7 @@
 PINK = "#FF00FF"
 TEAL = "paleturquoise3"
 
-block_letters = ["I", "L", "L_rev", "O", "E", "Z", "Z_rev"]
+block_letters = ["I"]
```
Akuli, line 24: We can't put this to main. Sorry.

tetris.py
```diff
@@ -88,9 +88,11 @@ def run_gui():
         None,
         None,
     )
-
     small_board.resize_to_fit()
 
+    print(small_board.block_letter)
+    print(small_board.current_block_center)
```
Akuli, line 94: Delete these

tetris.py
```diff
@@ -229,6 +230,7 @@ def resize_to_fit(self):
             self.current_block_center = (2, 0)
         else:
             self.current_block_center = (1, 1)
+        self.draw_block()
```
Akuli, line 233: This is confusing. I really wouldn't expect a function named `resize_to_fit` to draw anything to the canvas.

Akuli (changes requested):

Akuli (approved):
Includes some changes unrelated to fixing #55 but whatevs

PR #57, opened 2021-04-15
Revert "topbar_canvas now follows the shape of over new block EXCEPT for the …"

### LINKED ISSUE #55: 'I' block is not displayed in small_board

Akuli:
Reverts Arrinao/tetris#54 and fixes #55 

There's also #56, but it requires some reviewing whereas this PR is ready right now

If we merge this, we can later fix the code in #54 and make a new PR

### DIFF
--- a/tetris.py
+++ b/tetris.py
@@ -51,15 +51,8 @@ def run_gui():
     )
     topbar_score.pack(side="left", fill="x", expand=True)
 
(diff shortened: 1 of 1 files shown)


Akuli:
No need to revert, game still playable

PR #61, opened 2021-04-25
Attempt to make a new_game button start a new_instance of the game and delete the old one

Arrinao:


tetris.py
```diff
@@ -489,5 +468,28 @@ def timer(self):
             self.topbar_time.config(text=f"{int(game_time / 60):02d}:{int(game_time % 60):02d}")
             self.topbar_time.after(1000, self.timer)
 
+    def new_game(self, event):
```
Akuli, line 471: This can be called with event (from `.bind("<f>" ...)`) or without event (from `self.new_game()`):

```suggestion
    def new_game(self, event=None):
```

Akuli (changes requested):

tetris.py
```diff
@@ -165,8 +146,31 @@ def run_gui():
 
     root.title("Tetris – by The Philgrim, Arrinao, and Master Akuli")
     # root.iconphoto(False, tkinter.PhotoImage(file=image_name.png")) TODO: INSERT LATER
+
     root.mainloop()
 
+# Draws the board consisting of 15x10 squares before the game starts
```
Akuli, line 152: This should be a docstring, not a comment. But there's already a docstring below.

```suggestion
```

tetris.py
```diff
@@ -385,37 +365,37 @@ def full_line_clear(self):
         """
         Flashes the line once it's fully populated with blocks and clears it afterwards
         """
-        full_lines = []
+        self.full_lines = []
```
Akuli, line 368: Don't put it to `self`, it's weird and doesn't fix any problem

tetris.py
```diff
@@ -489,5 +467,28 @@ def timer(self):
             self.topbar_time.config(text=f"{int(game_time / 60):02d}:{int(game_time % 60):02d}")
             self.topbar_time.after(1000, self.timer)
 
+    def new_game(self, event=None):
```
Akuli, line 470: This never gets an event, because it's no longer used with bind and buttons don't give an event to their functions:
```suggestion
    def new_game(self):
```

Akuli (changes requested):

tetris.py
Arrinao: Removed, it was an attempt to fix the score

tetris.py
Arrinao: Removed

Akuli (approved):
Needs black

PR #62, opened 2021-04-29
Major rework of the classes

Arrinao:


tetris.py
```diff
@@ -60,6 +61,7 @@ def run_gui():
     topbar = tkinter.Frame(root, bg=D_GREY, relief="ridge")
     topbar.grid(row=0, columnspan=2, sticky="we")
 
+    global topbar_time
```
Akuli, line 64: Are these needed? Figure out which ones aren't, and delete them.

tetris.py
```diff
@@ -407,89 +467,65 @@ def flasher(self, full_lines, fill):
...
-        self.new_game()
+    def timer(self):
+        if self.game_status == GameStatus.in_progress:
+            game_time = time.time() - self.start_time - self.paused_time
+            self.topbar_time.config(text=f"{int(game_time / 60):02d}:{int(game_time % 60):02d}")
+            self.topbar_time.after(1000, self.timer)
 
-    def game_over_check(self):
-        y_coordinates = [y for (x, y) in self.main_board.coord_extractor()]
+    def game_over(self, forced=False):
```
Akuli, line 478: There are two `def game_over(...)` in this code, this one here in `class Game`, and another in `class TetrisControl`. The one in `Game` is never used. Please delete it.

tetris.py
```diff
@@ -407,89 +467,65 @@ def flasher(self, full_lines, fill):
...
-        self.game_status = GameStatus.in_progress
-        self.main_board.game_score = 0
-        self.main_board.topbar_score.config(text=self.main_board.game_score)
+    def game_over(self, forced=False):
+        y_coordinates = [y for (x, y) in self.game.coord_extractor()]
+        if any(y < 0 for y in y_coordinates):
+            self.game.game_status = GameStatus.game_over
+        elif forced == True:
+            self.game.game_status = GameStatus.game_over
+            new_game()
```
Akuli, line 526: You never want to run `new_game()` without setting the status of the old game to `game_over`. Setting the status should be a part of `new_game()`, so that it's impossible to forget it.

tetris.py
```diff
@@ -336,13 +395,14 @@ def user_input_right(self):
             return
         x, y = self.current_block_center
         self.current_block_center = (x + 1, y)
+        self.main_board.draw_block(self.get_block_shape(), self.block_letter, self.landed_blocks)
```
Akuli, line 398: This line of code is copy/pasted into many places. Maybe make a new method that contains it? Something like `def redraw(self): self.main_board....`

tetris.py
```diff
@@ -305,15 +357,21 @@ def move_current_block_down(self):
...
+                    self.landed_blocks[self.block_letter] = []
+                self.landed_blocks[self.block_letter].extend(self.get_block_shape())
+                self.full_line_clear()
+                self.game_over()
+                self.new_block()
+            elif not self.fast_down:
+                x, y = self.current_block_center
+                self.current_block_center = (x, y + 1)
+            self.main_board.draw_block(self.get_block_shape(), self.block_letter, self.landed_blocks)
+            self.main_board.canvas.after(game_speed, self.move_current_block_down)
```
Akuli, line 374: This can be simplified. Right now it's like:

```
if self.game_status == GameStatus.paused:
    self.main_board.canvas.after(...)
elif self.game_status == GameStatus.in_progress:
    ...lots of code...
    self.main_board.canvas.after(...)
```

This would be simpler:

```
if self.game_status == GameStatus.in_progress:
    ...lots of code...

if self.game_status != GameStatus.game_over:
    self.main_board.canvas.after(...)
```

tetris.py
```diff
@@ -296,6 +319,35 @@ def get_block_shape(self):
...
+            coords = [(x - 1, y - 1), (x - 1, y), (x, y), (x + 1, y)]
+
+        if self.upcoming_block_letter == "O":
+            coords = [(x - 1, y), (x, y), (x, y - 1), (x - 1, y - 1)]
+
+        if self.upcoming_block_letter == "E":
+            coords = [(x - 1, y), (x, y), (x + 1, y), (x, y - 1)]
+
+        if self.upcoming_block_letter == "Z":
+            coords = [(x - 1, y - 1), (x, y - 1), (x, y), (x + 1, y)]
```
Akuli, line 345: There's no need to create a `coords` variable. Just return the lists directly. For example:

```suggestion
        if self.upcoming_block_letter == "Z":
            return [(x - 1, y - 1), (x, y - 1), (x, y), (x + 1, y)]
```

tetris.py
```diff
@@ -296,6 +319,35 @@ def get_block_shape(self):
 
         return coords[self.rotate_counter % len(coords)]
 
+    def get_upcoming_block_shape(self):
+        if self.upcoming_block_letter == 'I':
+            self.upcoming_block_center = (2, 0)
+        else:
+            self.upcoming_block_center = (1, 1)
+        (x, y) = self.upcoming_block_center
```
Akuli, line 327: There's no need to store the center to `self`. If you use a local variable instead, the name can be shorter too.

```suggestion
        if self.upcoming_block_letter == 'I':
            center = (2, 0)
        else:
            center = (1, 1)
        (x, y) = center
```

tetris.py
```diff
@@ -220,44 +223,64 @@ def draw_block(self):
...
-            self.current_block_center = (1, 1)
+
+
+class Game:
+    def __init__(self, main_board, small_board, topbar_score, topbar_time):
+        self.landed_blocks = {}
+        self.block_letter = random.choice(block_letters)
+        self.upcoming_block_letter = random.choice(block_letters)
+        self.current_block_center = (int(game_width / 2), -2)
+        self.upcoming_block_center = (1, 1)
```
Akuli, line 262: Never used.

```suggestion
```

tetris.py
```diff
@@ -220,44 +223,64 @@ def draw_block(self):
...
+        self.topbar_score = topbar_score
+        self.main_board = main_board
+        self.small_board = small_board
+        self.topbar_time = topbar_time
+        self.fast_down = False
+        self.start_time = time.time()
+        self.rotate_counter = 0
+        self.game_score = 0
```
Akuli, line 270: Rename `game_score` to `score`. It's inside `class Game`, so there's no need to repeat `game` in front of it.

tetris.py
```diff
@@ -220,44 +223,64 @@ def draw_block(self):
...
-        if self.block_letter == "I":
+        if block_letter == "I":
             self.canvas.config(width=square_size * 4, height=square_size * 1)
-
-        if self.block_letter == "I":
             self.canvas.pack(pady=square_size / 2 + 10)
-            self.current_block_center = (2, 0)
         else:
             self.canvas.pack(pady=10)
-            self.current_block_center = (1, 1)
```
Akuli, line 260: I think the logic was more clear before. Are you sure you don't want a separate resizing part (everything that uses `self.canvas.config`) and padding part (everything that uses `self.canvas.pack`)?

tetris.py
```diff
@@ -205,11 +208,11 @@ def draw_rectangle(self, x, y, tags, fill):
             fill=fill,
         )
 
-    def draw_block(self):
+    def draw_block(self, block, block_letter, landed_blocks):
```
Akuli, line 211: `landed_blocks` not needed for `small_board`, so add handy default:

```suggestion
    def draw_block(self, block, block_letter, landed_blocks=None):
```

tetris.py
```diff
@@ -220,44 +223,64 @@ def draw_block(self):
...
+        self.paused_time = 0
+        self.game_status = GameStatus.in_progress
+        self.timer()
+
+    def new_block(self):
+        if self.game_status == GameStatus.in_progress:
+            self.current_block_center = (int(game_width / 2), -2)
+            self.block_letter = self.upcoming_block_letter
+            self.upcoming_block_letter = random.choice(block_letters)
+            self.small_board.draw_block(self.get_upcoming_block_shape(), self.upcoming_block_letter, self.landed_blocks)
```
Akuli, line 281: landed blocks not needed for small board:

```suggestion
            self.small_board.draw_block(self.get_upcoming_block_shape(), self.upcoming_block_letter)
```

tetris.py
```diff
@@ -175,6 +176,15 @@ def draw_board(canvas):
         x_gap += square_size
 
 
+def new_game():
+    small_board = Board(topbar_canvas, None)
+    main_board = Board(game_canvas, small_board)
+    game = Game(main_board, small_board, topbar_score, topbar_time)
+    game.move_current_block_down()
+    small_board.draw_block(game.get_upcoming_block_shape(), game.upcoming_block_letter, game.landed_blocks)
```
Akuli, line 184: landed blocks not needed for small board:
```suggestion
    small_board.draw_block(game.get_upcoming_block_shape(), game.upcoming_block_letter)
```

Akuli (changes requested):

Akuli:
Stuff to do in later reviews, easier to do once other review problems are fixed:
- [ ] `get_block_shape()` copy/pasta cleanup
- [ ] getting rid of `Board.small_board`

tetris.py
Akuli: Maybe I should clarify what I meant.

Right now the code is like:

```
        ...
        if block_letter == "Z_rev":
            self.canvas.config(...)
        if block_letter == "I":
            self.canvas.config(...)
            self.canvas.pack(pady=square_size / 2 + 10)
        else:
            self.canvas.pack(pady=10)
```

Separating `config` and `pack` parts would make it easier to read:

```
        ...
        if block_letter == "Z_rev":
            self.canvas.config(...)
        if block_letter == "I":
            self.canvas.config(...)

        if block_letter == "I":
            self.canvas.pack(pady=square_size / 2 + 10)
        else:
            self.canvas.pack(pady=10)
```

tetris.py
Akuli: something like

```
if tetris_control.game is not None:
    tetris_control.game.status = GameStatus.game_over
```

tetris.py
Akuli: Maybe `redraw_main_board` would be a better name. Choose something that makes sense to you.

tetris.py
Arrinao: Hmm, from what I can see, all of them are needed.

tetris.py
Akuli: This comment still applies, even after the copy/pasta cleanup.

tetris.py
```diff
@@ -220,81 +225,113 @@ def draw_block(self):
...
+            self.upcoming_block_center = (2, 0)
+        else:
+            self.upcoming_block_center = (1, 1)
+        (x, y) = self.upcoming_block_center
 
-        if self.block_letter == "Z":
-            coords = [[(x - 1, y - 1), (x, y - 1), (x, y), (x + 1, y)]]
-            coords.append([rotate_point(point, self.current_block_center) for point in coords[-1]])
+        coords = self.get_coords(self.upcoming_block_letter, x, y)
+        return coords
```
Akuli, line 310:
```suggestion
        return self.get_coords(self.upcoming_block_letter, x, y)
```

Akuli (changes requested):

Akuli (changes requested):

tetris.py
Arrinao: Resolved as part of get_block_shape change

tetris.py
```diff
@@ -220,82 +217,120 @@ def draw_block(self):
         }
         self.canvas.delete("block")
 
-        for x, y in self.get_block_shape():
-            self.draw_rectangle(x, y, "block", self.color_dict[self.block_letter])
-
-        for letter, coord_list in self.landed_blocks.items():
-            for (x, y) in coord_list:
-                self.draw_rectangle(x, y, "block", self.color_dict[letter])
+        if self.small_board is not None:
```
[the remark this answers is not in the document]
Akuli: It is still not a boolean! Its still a board or `None`. Also it should be renamed to `is_main_board` or something.

tetris.py
Akuli: Please do this instead of just marking it resolved!

Akuli (changes requested):

Akuli (approved):

PR #64, opened 2021-05-11
separate config and pack parts of code

Akuli:
Somehow arrinao hates this, and didn't do this in #62. He still hasn't explained to me why he wants to have the `config` and `pack` parts of the code squeezed together in a weird way that doesn't make any sense.


Arrinao:
It looks weird to have two conditions for the same thing, but I'll do this if you think it's better this way.

Akuli:
Basically, the `pack` handling for `"I"` is special, and that should not be hidden by making it similar to other lines of code. But that said, it's your project; you decide whether you approve or close this PR.

Arrinao (approved):

PR #67, opened 2021-05-12
Added a dialogue window for verifying the wish of the user to end a current game

Arrinao:
This also pauses the game.

### DIFF
--- a/tetris.py
+++ b/tetris.py
@@ -195,7 +195,7 @@ def start_dialogue():
         if mb.askokcancel('End current game?', 'Do you want to end the current game and start anew?', parent=topbar_time):
             new_game()
         else:
-            tetris_control.pause_game()
+            return
 
     else:
         new_game()
--- a/tetris.py
+++ b/tetris.py
@@ -179,7 +179,7 @@ def draw_board(canvas):
 
 def new_game():
     if tetris_control.game is not None:
-        tetris_control.game.game_status = GameStatus.game_over
+        tetris_control.game.status = GameStatus.game_over
     small_board = Board(topbar_canvas, False)
     main_board = Board(game_canvas, True)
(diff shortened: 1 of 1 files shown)


tetris.py
```diff
@@ -187,6 +188,17 @@ def new_game():
     tetris_control.game = game
 
 
+def start_dialogue():
+    if tetris_control.game.game_status == GameStatus.in_progress or tetris_control.game.game_status == GameStatus.paused:
+        tetris_control.pause_game()
+        if mb.askokcancel('End current game?', 'Do you want to end the current game and start anew?', parent=topbar_time):
+            new_game()
+        else:
+            tetris_control.pause_game()
```
Akuli, line 197: I don't think this logic works if the game is already paused when you click new game button.

tetris.py
```diff
@@ -187,6 +188,23 @@ def new_game():
...
+        if tetris_control.game.status != GameStatus.paused:
+            game_running = True
+            tetris_control.pause_game()
+        else:
+            game_running = False
+
+        if mb.askokcancel('End current game?', 'Do you want to end the current game and start anew?', parent=topbar_time):
+            new_game()
+        else:
+            if game_running is True:
```
Akuli, line 202:
```suggestion
            if game_running:
```

tetris.py
```diff
@@ -187,6 +188,23 @@ def new_game():
     tetris_control.game = game
 
 
+def start_dialogue():
```
Akuli, line 191: Consider renaming this. A function named `start_foo()` gives me the impression that the function ends before `foo` completes. But this function waits for the dialogue to complete, because that's how `mb.askokcancel` works.

Akuli (commented):
Looks good overall! A couple small nits.

Akuli (approved):
Black and we can merge

PR #68, opened 2021-05-12
Added a condition for move_block_down_press on TetrisControl

Arrinao:
Quick fix for bug: when paused, block could be moved down

### DIFF
--- a/tetris.py
+++ b/tetris.py
@@ -509,8 +509,7 @@ class TetrisControl:
             self.game.user_input_right()
 
     def move_block_down_press(self, event):
-        if self.game.status == GameStatus.in_progress:
-            if not self.game.fast_down:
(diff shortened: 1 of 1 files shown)


tetris.py
```diff
@@ -509,9 +509,10 @@ def move_block_right(self, event):
             self.game.user_input_right()
 
     def move_block_down_press(self, event):
-        if not self.game.fast_down:
-            self.game.fast_down = True
-            self.game.user_input_down()
+        if self.game.status == GameStatus.in_progress:
+            if not self.game.fast_down:
```
Akuli, line 513:
```suggestion
        if self.game.status == GameStatus.in_progress and not self.game.fast_down:
```

Akuli (changes requested):

Akuli:
black

Akuli (approved):

PR #70, opened 2021-05-19
High_scores window will show elapsed time, game speed and score of the past games. 

Arrinao:
Sorted by the scores.


tetris.py
```diff
@@ -39,6 +40,16 @@
     # this file is.
     image_dir = pathlib.Path(__file__).parent / "images"
 
+try:
+    with open(image_dir / "game_data.json", "r") as source:
+        json_dict = json.load(source)
+except FileNotFoundError:
+    # TODO: game_data.json shouldn't be in image_dir
```
Akuli, line 47: Fix this instead of copying bad code around.

tetris.py
```diff
@@ -39,6 +40,16 @@
...
+    json_dict = {
+        "high_scores": [],  # list of dicts with keys: 'time', 'width', 'height', 'mine_count'
```
Akuli:
```suggestion
        "high_scores": [],  # list of dicts with keys: 'Time', 'Game Speed', 'Score'
```

tetris.py
```diff
@@ -282,6 +293,7 @@ def __init__(self, main_board, small_board, topbar_time):
         self.topbar_time = topbar_time
         self.fast_down = False
         self.start_time = time.time()
+        self.time = 0
```
Akuli, line 296: Same data in two places. Please no. Make a function that calculates and returns the time.

Akuli (commented):
Much work left to do but here's review so far

tetris.py
Akuli: Actually, ThePhilgrim fixed this already on minesweeper, but he's not responding to my PR review on minesweeper. The correct code is in his PR.

Arrinao: Deleted try part for now

tetris.py
```diff
@@ -475,16 +482,21 @@ def flasher(self, full_lines, fill):
...
                 self.main_board.draw_rectangle(x, x_line, "flash", fill)
 
+    def get_time(self):
+        return time.time() - self.start_time - self.paused_time
+
     def timer(self):
         if self.status == GameStatus.in_progress:
-            game_time = time.time() - self.start_time - self.paused_time
-            self.topbar_time.config(text=f"{int(game_time / 60):02d}:{int(game_time % 60):02d}")
+            self.get_time()
```
Akuli, line 490: What is this line of code supposed to do? It calls the function without using the return value for anything.

Akuli:
I tried the game, and there are some problems.

- [ ] High score window doesn't go away when I click "New Game"
- [ ] If I click "High Scores" while the game is paused, it unpauses
- [ ] If I click "High Scores" while the game is running, it doesn't unpause when I close high scores window
- [ ] There's no "Back to game" button. You have to click "High Scores", which is not what you expect; instead of displaying high scores (which you're already looking at), it goes back to the game.
- [ ] High scores are lost when I restart the game

Also, the highscore list looks bad:

hs

- [ ] Custom background color is not working when there is something in the treeview
- [ ] "Time Spent" numbers are displayed too accurately, needs rounding
- [ ] "Game Speed" column is too narrow
- [ ] "Score" column is too wide

Arrinao:
Problem #3 with no unpausing when high_scores dissapears is intentional.

tetris.py
```diff
@@ -221,6 +225,9 @@ def display_highscores():
 def new_game():
     if tetris_control.game is not None:
         tetris_control.game.status = GameStatus.game_over
+    if treeview:
```
Akuli, line 228: I don't think this works. Do you want to say "if the treeview is packed"? If you do, make a function that returns `True` when a widget is packed and `False` when it isn't, using the code you already wrote for this.

Arrinao: The code works!

tetris.py
```diff
@@ -117,12 +129,30 @@ def run_gui():
     global tetris_control
     tetris_control = TetrisControl()
 
+
+    global treeview
+    treeview = ttk.Treeview(game_frame, columns=('Time Spent', 'Game Speed', 'Score'), height=23)
+    treeview.tk.eval('''ttk::style theme use clam
+    ttk::style configure Treeview -fieldbackground gray7 -bordercolor royalblue''')
```
Akuli, line 136: Would be more readable like this:

```suggestion
    treeview.tk.eval('''
    ttk::style theme use clam
    ttk::style configure Treeview -fieldbackground gray7 -bordercolor royalblue
    ''')
```

Akuli:
High score list still looks kinda bad. Is it better with the correct font?

hugefont

Arrinao:
Yes. The 'Arrr Matey' font is smaller than the default one.

Arrinao:
image

tetris.py
```diff
@@ -177,9 +228,46 @@ def draw_board(canvas):
         x_gap += square_size
 
 
+def get_high_scores():
```
Akuli, line 231: This function needs to be renamed. How about `update_high_score_view`?

tetris.py
```diff
@@ -177,9 +228,46 @@ def draw_board(canvas):
...
+        index += 1
+        index %= 2
+    treeview.tag_configure('odd_row', background=D_GREY)
+    treeview.tag_configure('even_row', background='black')
+
+
+def display_high_scores():
+    get_high_scores()
+    try:
+        game_canvas.pack_info()
```
Akuli, line 249:
```suggestion
        game_canvas.pack_info()  # Raises error if canvas not packed
```

tetris.py
```diff
@@ -177,9 +228,46 @@ def draw_board(canvas):
...
+        game_canvas.pack_forget()
+        if tetris_control.game.status == GameStatus.in_progress:
+            tetris_control.pause_game()
+        treeview.pack(side='top', fill='both', expand=True)
+    except tkinter.TclError:
```
Akuli, line 254: This is supposed to catch `tkinter.TclError` raised from `game_canvas.pack_info()`. It does that, but it also catches `tkinter.TclError` raised from the rest of the code inside `try`. Look up "python try except else" and use that to clean this up.

tetris.py
```diff
@@ -177,9 +228,46 @@ def draw_board(canvas):
...
+        treeview.pack_forget()
+        game_canvas.pack(fill='both', expand=True)
+
+
+def clear_high_scores():
```
Akuli, line 259: It's too easy to clear the high scores accidentally. Maybe ask the user whether they really want to clear them, similar to starting a new game?

tetris.py
```diff
@@ -49,23 +60,29 @@ def set_button_image(button_image, event):
 
 
 def run_gui():
+    global game_frame
```
Akuli, line 63: Is the globalling needed?

tetris.py
```diff
@@ -177,9 +228,46 @@ def draw_board(canvas):
...
+    json_dict['high_scores'] = []
+    with open('game_data.json', 'w') as game_data:
+        json.dump(json_dict, game_data)
+
+
 def new_game():
     if tetris_control.game is not None:
         tetris_control.game.status = GameStatus.game_over
+    treeview.pack_forget()
+    game_canvas.pack(fill='both', expand=True)
```
Akuli, line 270: In this context, it's not clear what `treeview` is. Please rename it to `high_scores_treeview` or something.

tetris.py
```diff
@@ -475,16 +563,22 @@ def flasher(self, full_lines, fill):
...
-            game_time = time.time() - self.start_time - self.paused_time
-            self.topbar_time.config(text=f"{int(game_time / 60):02d}:{int(game_time % 60):02d}")
+            self.topbar_time.config(text=self.get_time())
             self.topbar_time.after(1000, self.timer)
 
     def game_over(self):
         y_coordinates = [y for (x, y) in self.coord_extractor()]
         if any(y < 0 for y in y_coordinates):
             self.status = GameStatus.game_over
+            json_dict['high_scores'].append({'Time': self.get_time(), 'Game Speed': game_speed, 'Score': self.score})
```
Akuli, line 579: The JSON contains times as strings, making it difficult to sort the contents of the treeview, in case we want to do that later.

tetris.py
```diff
@@ -177,9 +228,46 @@ def draw_board(canvas):
...
 
+def get_high_scores():
+    index = 0
+    treeview.delete(*treeview.get_children())
+    for high_score_dict in json_dict['high_scores']:
+        if index == 0:
+            tag = 'odd_row'
+        else:
+            tag = 'even_row'
+        treeview.insert(parent="", index="end", tags=tag, values=(high_score_dict['Time'], high_score_dict['Game Speed'], high_score_dict['Score']))
```
Akuli, line 239: The speed shows up in the treeview as "300". It's not very useful. Showing it as "300ms/block" would be much better.

Akuli (changes requested):

Akuli:
Looking at the screenshot you posted while I was reviewing, the game looks just epic. Good job!

tetris.py
```diff
@@ -117,26 +138,50 @@ def run_gui():
     global tetris_control
     tetris_control = TetrisControl()
 
+    style = ttk.Style(game_frame)
+    style.theme_use('default')
+    style.configure('Treeview', rowheight=25, background=D_GREY, foreground='dark orange', fieldbackground=D_GREY, font=('Bahnschrift Condensed', 16))
+    style.map('Treeview', background=[('selected', '#BFBFBF')], foreground=[('selected', 'black')])
+    style.configure("Treeview.Heading", background='gray4', foreground='orangered', font=('Arrr Matey BB', 22), padding=[5,0])
```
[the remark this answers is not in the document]
Arrinao: Arrr Matey BB is THE font name. The font file is named without the 'BB' I think.

tetris.py
```diff
@@ -177,9 +228,46 @@ def draw_board(canvas):
...
+    index = 0
+    treeview.delete(*treeview.get_children())
+    for high_score_dict in json_dict['high_scores']:
+        if index == 0:
+            tag = 'odd_row'
+        else:
+            tag = 'even_row'
+        treeview.insert(parent="", index="end", tags=tag, values=(high_score_dict['Time'], high_score_dict['Game Speed'], high_score_dict['Score']))
+        index += 1
+        index %= 2
```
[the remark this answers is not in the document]
Arrinao: Done, thanks for the suggestion. I knew there was an easier way to handle this

tetris.py
Arrinao: changed to get_high_scores_into_treeview

tetris.py
```diff
@@ -1,18 +1,20 @@
...
 from enum import Enum
 from tkinter import messagebox as mb
+from tkinter import ttk
 
 game_speed = 300
 square_size = 32
 game_width = 10
 game_height = 15
-sidebar_width = 100
+sidebar_width = 106
```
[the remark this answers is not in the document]
Arrinao: Changed back to 100

tetris.py
Arrinao: I think it would be better to just display a difficulty name instead of number value and make it a primary sorting tool

tetris.py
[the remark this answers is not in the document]
Akuli: This PR includes only one font file, `bahnschrift.ttf`.

tetris.py
Arrinao: Added a dialogue window for clear list

tetris.py
Arrinao: I'll do this in another PR though.

README.md
```diff
@@ -1,2 +1,21 @@
...
+
+The program right now only features the game Tetris. It will feature more games in the future.
+
+There is a High Scores list avaliable to access from the menu on the right side, that displays the best achievements in the games.
+
+The game itself is controllable with a keyboard. Pressing P will pause the game; arrow keys will move the blocks around. The Up arrow key in tetris serves as the rotate button.
+
+**Installation info:**
+
+All the font files (.ttf) need to be moved into the Fonts folder. On windows the directory is C:\Windows\Fonts.
```
Akuli, line 13:
```suggestion
All the font files (.ttf) need to be copied into the Fonts folder. On windows the directory is C:\Windows\Fonts.
```

tetris.py
Akuli: Can't see the code for it?

Akuli (commented):
Seems like you haven't pushed in a while?´

Akuli:
Weird. The review appeared as if I was adding more comments to existing conversations. I edited the comments so they make sense without the previous conversations.

tetris.py
Arrinao: Didn't push :( Damn it!

Akuli (approved):
Darken the n-word and we can merge :)

PR #71, opened 2021-06-10
Fix font list formatting

Akuli:
Currently README looks pretty bad on github. Second font shows up differently from first. This fixes.


Arrinao (approved):
Thanks, this was needed

PR #74, opened 2021-09-27
screenshot in readme

Arrinao:


Akuli (approved):
The alt text is a bit weird, but whatever :)

